PR #10, opened 2017-11-14
First unit test and beginning framework

AricStewart:
Validate allocation and divisional data


AricStewart:
Unnecessary changes pulling things out of functions.inc into config.inc  undoing that.

PR #25, opened 2017-11-26
Expand user search to include Name

AricStewart:


tskeeley:
That's part of the reason for all of the front end.  Duplicates are a huge problem.  About 1/4 of the total account list is duplicate or failed info.  The find duplicate function in Neon is... lacking.  Something as simple as not creating blank accounts for "guest of" badges or verifying that every email address is unique, would go a very long way to cleaning up most of the issues.  It would also limit the number of accounts a company has in neon, and Z2Systems charges by count of accounts in the database.  So, it's not really in their best interests to fix the duplicate account issue, even though it remains the number one request from their clients. (to verify unique email, I mean)

Front ending all of this is a first step to moving to a non-vendor dependant backend allowing us to stay with Neon OR leave them.  The choice is a big part of keeping things sustainable.  And that's why we are trying to move to abstracting the data storage from the tool used.  (How is that for too much of an explanation to a simple comment? :) )

AricStewart:
Sounds like you could open an issue or issues around this to better track what needs to be done! :)

tskeeley:
LOL, good reminder  will do

PR #35, opened 2017-12-04
Add warning for overlapping hours

AricStewart:


AricStewart:
That is correct. I had not thought through that I will rework and examine. Make it more robust.

PR #37, opened 2017-12-05
Aric pull thing out of neon

AricStewart:


AricStewart:
Ok Dropping the doGeneralSearch part of the change.

PR #40, opened 2017-12-10
Aric neon cleanup

AricStewart:
Continue pulling Neon specific things out of the web page templates and into functions


functions/session.inc
```diff
@@ -4,7 +4,7 @@
 function buildSessionData($accountId)
 {
   // Refresh Account info from Neon to verify update
-    require_once($FUNCTIONDIR."/users.inc");
+    require_once(dirname(__FILE__)."/users.inc");
```
tskeeley: This feels odd, setting it to not use the common variable that is used everywhere else.  But I see the point.  Does changing this actually buy us anything or just a preference?

AricStewart: It is pretty common PHP coding to do this sort of thing. since the two files are in the same directory. 

It also lets my phplint tool find the included file and be able to more sanely check my code as I develop.

tskeeley: Very cool.  Thanks!

PR #45, opened 2018-01-21
Work to display volunteer hours and prizes

AricStewart:
Note there is a debug tool I have added
test/populate_volunteer.php

That php script will **WIPE OUT EXISTING** volunteer data (hours, prizes, claims, etc...) and populate the database with a bunch of random data.  That should help visualization of the work. 

Next step is work to allow selection of prizes.

### DIFF
--- a/functions/database.inc
+++ b/functions/database.inc
@@ -1,5 +1,9 @@
 <?php
 
+/*.
+    require_module 'standard';
+.*/
+
 // Currently, this database class is set for the MySQL PDO filter/backend. Future version will expand it to be agnostic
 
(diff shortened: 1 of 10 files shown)


AricStewart:
Ok so how do we to do this?
Likely we just need add a field somewhere that specify current year. It would also give us the nice opportunity to be able to view/edit previous years as well.

So do we want current year to be in the `Configurations` table in the DB or in the local configuration file?   I would say the `Configurations` table and maybe call it `CurrentYearID`?

-aric

AricStewart:
How does that look?

tskeeley:
That will work!  Merging

PR #87, opened 2018-03-10
Database 2018031000, includes fixes caused by relationships imposed

tskeeley:
Update DB to 2018031000


AricStewart:
Looks good! Thanks!

PR #90, opened 2018-03-24
Tk database schema2018031200

### LINKED ISSUE #88: Merge Year data with event data
Merge Year data into event data to avoid duplicated data efforts

tskeeley:
@AricStewart This is the DB update you asked for a while ago.  Fixes #88 

ONE NOTE:  It's possible that the update will grump about contraints since the yearID and EventID are all blank.  I did a manual update of the eventid in concomlist to update to eventID 20, and applied the concom constraint manually because I wasn't sure what the issue was.  I don't think it would be a real problem during an install, only a problem because we had all that test data in the db.


tskeeley:
interesting.

So, the current_yearID...   I don't know what that is about, but I'll hunt it.

The yearID one that comes from from_neon_to_152 is happening because we got rid of the yearID table and fields when we merged.  so, we need to update the pull from neon routine.  I can look at that later as well

AricStewart:
We have not released yet and we have proven our upgrade path to some point. I think we have 2 options. 

1) require fresh reimport and rewrite our neon_to_xxx function to put things in the right places. 

2) do the table changes via an upgrade here instead of in the base tables  

I am perfectly happy with #1 both mean thoughtful work. 

-aric

AricStewart:
Getting further. Making proper changes to current_yearID ( becoming current_cycleID) and fixing up may other places we are specifying a YearID. 

I am unable to log into Neon with my instance anymore though. I get “Api key is invalid” when I try.

AricStewart:
Ok I have push all the around the seams work to make this database change work. 

Please review and we can push this.

PR #113, opened 2018-03-30
Tk db activity log

tskeeley:
Update to capture (non-Select) queries.  Also display on the admin page.


AricStewart:
Looks great! Thanks!

PR #120, opened 2018-04-03
Cleanup Volunteer hour rounding

AricStewart:
Also allow the rounding value to be easily changed and customized


AricStewart:
Fully agree! I just was not in the mood to modify configuration yet.  We can take that step if you like. Seems like a fair one.

tskeeley:
no, let's save it for the cleanup round


-- Thomas Keeley

If a Smurf holds its breath for too long, what color does it turn??
Para. Jeff Dunham

AricStewart:
understood!  BTW check slack. Have a question for you.

PR #133, opened 2018-04-06
use jscs to validate javascript code

AricStewart:


AricStewart:
Further commit coming with jshint also

Issue #80, opened 2018-03-01
Registration Worker: First/Last Name Lookup


seanbseanb:
As a Registration Worker, I want to be able to look up a guest's event registration records by their last name or first name.

I want to be able to see:

1. How many badges they purchased
2. The names associated with those badges

seanbseanb:
Where do I see this lookup page?

PR #148, opened 2018-04-15
Add event registration badge lookup page

AricStewart:
Potentially for 2019? Or would we use it for 2018?


AricStewart:
For Issue #80

Issue #83, opened 2018-03-01
Registration Guest: Reset Password


seanbseanb:
As a Registration Guest, I want to be able to reset the password for my login.  This should use an existing, registered email or SMS as password reset options.

AricStewart:
This work is present in the `ARIC_remove_Neon` branch for examination.

AricStewart:
This functionality is in place

Issue #81, opened 2018-03-01
Registration Guest: Authenticate using existing credentials


seanbseanb:
As a Guest, I want to be able to authenticate to CIAB using credentials I already have.

If I have forgotten my password, I want to be able to reset it from the authentication screen.

tskeeley:
The user already does authenticate using credentials.  All users do.  Is this like #82 but for password change/recovery without logging into neon?

seanbseanb:
How will we authenticate post-neon?  The idea here is to use another authentication system so that we can work without neon.  Can we rough in some auth tables and setup our own authentication?

AricStewart:
I have implemented a basic User/Password login. the Username(or email) stays the same from Neon days but the user will need to reset their password.

Or if we have some overlap time we could fake populate the database by checking a given password vs Neon and populating our password file on success.

AricStewart:
This has been in place since 2018.

PR #172, opened 2018-06-11
Migrate Passwords from Neon

AricStewart:
Migrate user password as they are used from Neon to our local Database.

This will make abandoning Neon much less painful for the user base.

Addresses Issues #83 and #81


AricStewart:
Sorry for the noise in the above commits, rebasing and correcting an error mid sequence causes the output to be confusing.

PR #182, opened 2018-07-05
Allow Volunteer hours to be less than 1

AricStewart:


cvg-it:
FYI, I have merged this in the interest of expediency. 

--
Sean Berry Co-Head / IT
it@convergence-con.org
414-339-1033

PR #186, opened 2018-07-09
Move away from Neon for states and countries

AricStewart:


tskeeley:
Is this something we need to merge right now?  Can it wait until after the vol party to minimize changes?

AricStewart:
It can wait. Should be harmless but also no need for risk.

PR #200, opened 2018-09-15
format cleanup for CFO Staff - static entry for now

tskeeley:
This is mostly about supporting the separation of CFO from the Systems Division.


tskeeley:
Fair catch.  I want this part moved to configuration, but that's a cleanup thing for later.  Updated now and you can review

PR #237, opened 2019-02-04
Update DumpConComList to include aliases

tskeeley:
aliased departments are getting dropped from the google group because the aliases will SOMETIMES show before the department's main email.  I opted for this version instead of calling "not alias" because some of these groups are not the same users in the end.  AFTER the email fix on the google apps domain, this might need to switch to no-alias, but I don't want to decide that until we know the final form of the email domain


cvg-it:
??

--
Sean Berry Co-Head / IT
it@convergence-con.org
414-339-1033

PR #247, opened 2019-03-09
Further work to aid in duplicate account handling

AricStewart:
Check credentials agenst ALL the accounts to let the user in if
the password matches any of the accounts with the email

Also if everything fails display a warning at the login to help
informe the user that there is a possible issue

### DIFF
--- a/functions/authentication.inc
+++ b/functions/authentication.inc
@@ -92,6 +92,11 @@ function checkNeonForUserAndPasswordMatch($user, $password)
 
 }
 
+const AUTH_SUCCESS   = 0x0;
+const AUTH_BAD       = 0x1;
+const AUTH_EXPIRED   = 0x2;
+const AUTH_LOCKED    = 0x3;
+const AUTH_DUPLICATE = 0x10;
 
 function check_authentication($account, $password, $user)
 {
@@ -110,7 +115,7 @@ SQL;
     $value = $result->fetch();
     if ($value !== false) {
         if ($value['FailedAttempts'] >= $max_fail) {
-            return -1;
+            return AUTH_LOCKED;
         }
         if (password_verify($password, $value['Authentication'])) {
             if ($value['OneTime']) {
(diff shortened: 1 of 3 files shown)


mshappe:
How painful would it be (and how advisable) to, on any SUCCESSFUL authentication against ANY account, pull ALL of the duplicates and simply present them as if they were one account? I'm prepared to be talked out of the idea entirely if we think it's insecure.

This PR looks good to me but again with my PHP being a bit rusty so I'll save formal approval for @tskeeley

AricStewart:
> How painful would it be (and how advisable) to, on any SUCCESSFUL authentication against ANY account, pull ALL of the duplicates and simply present them as if they were one account? I'm prepared to be talked out of the idea entirely if we think it's insecure.
> 

This commit does the first part of this.  It compares the given credentials to ALL the accounts with that e-mail address and lets you into the one that matches if any and then warns about the rest. 

I am thinking we may use this for minor without an e-mail... That way they can have a separate account but not have to have a separate e-mail.  Then the one who they are dependent on can access their dependents accounts.  (Still in the planing works)

If we find 2 accounts without a dependency relationship then we would display a way for the user to handle the merge or defer to an admin in the case of NEON. 

> This PR looks good to me but again with my PHP being a bit rusty so I'll save formal approval for @tskeeley

Ok cool.

mshappe:
> This commit does the first part of this. It compares the given credentials to ALL the accounts with that e-mail address and lets you into the one that matches if any and then warns about the rest.

And that's definitely a big help! I was just looking for ways to ease the responsibility for even knowing there **were** duplicates away from the user. Given that a lot of these duplicates were created inadvertently, as I understand it, it might be worthwhile to simply sweep them under the rug, as it were. 

> I am thinking we may use this for minor without an e-mail... That way they can have a separate account but not have to have a separate e-mail. Then the one who they are dependent on can access their dependents accounts. (Still in the planing works)

I just responded to the whole dependency question in our email thread. There's clearly two questions here: what we do to smooth things out while we're still using Neon, and what we do in our own system where we get to design it according to our needs :-)

tskeeley:
Once we are away from Neon, I REALLY like the idea of auto-squishing
accounts, especially during import.  We'll need an Admin way to un-squish
in the case of "WHAT THE ... did you do?!", but that's going to be rare to
very rare, I would think.

I'll jump to git and review the pull now.

-- Thomas Keeley

If a Smurf holds its breath for too long, what color does it turn??
Para. Jeff Dunham

PR #249, opened 2019-04-03
Use composer and remove static library

AricStewart:


mshappe:
You should add `library` (I think) to `.gitignore` and then `git rm --keep` library and commit. You don't want to actually commit the library bits.

mshappe:
Oh, wait! NEVER MIND. Those are deletions of the OLD library.

AricStewart:
> Oh, wait! NEVER MIND. Those are deletions of the OLD library.

Exactly I am removing the library. :)

PR #250, opened 2019-04-04
Initial docker dev stuff

mshappe:
`Dockerfile.dev` will build a simple PHP-Apache docker environment. It will **not** copy in the source tree (the way Bryce's Dockerfile will) because...

`docker-compose.yml` will bring up mysql 5.7 and the app's docker image and link them, and grab port 80. Everything should more or less behave as if it were local, with one exception: the `.ht_blahblah` file for some reason is getting created with root perms despite Apache running as user 1000 (on the assumption that your logged in Linux user is likely user 1000). This can be easily fixed with a `sudo chown` and at some point I'll figure out how to make it not happen.

The one other thing you need to do now is run `composer install` locally. My next step is to figure out a clever way to have docker-compose run a startup script in the image that will do this if it's necessary, and only then proceed to start up Apache.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+Dockerfile*
+docker-compose*
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,4 @@
+FROM php:apache
+RUN apt update && apt install -y libcurl4-openssl-dev
+RUN a2enmod headers
+RUN docker-php-ext-install curl pdo pdo_mysql mysqli
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,4 +1,18 @@
 FROM php:apache
-RUN apt update && apt install -y libcurl4-openssl-dev
+ENV APP_DIR=/var/www/html
+
+RUN useradd -u 1000 ciab
+RUN apt update && apt install -y libcurl4-openssl-dev libzip-dev zlib1g-dev git
 RUN a2enmod headers
-RUN docker-php-ext-install curl pdo pdo_mysql mysqli
+RUN docker-php-ext-install curl pdo pdo_mysql mysqli zip
+#Install Composer
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+RUN php composer-setup.php --install-dir=. --filename=composer
+RUN mv composer /usr/local/bin/
+
+VOLUME ${APP_DIR}
+WORKDIR ${APP_DIR}
+COPY --chown=ciab . /var/www/html
+COPY docker-entrypoint.sh /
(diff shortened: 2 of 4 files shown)


mshappe:
Also, the passwords etc in the docker-compose file are dummies for now but they do have to match what you plan to use when you run `configure_system.php`

AricStewart:
So this docker files does not quite set everything up...
Any chance docker files like this could be an inspiration? https://github.com/alextselegidis/easyappointments/tree/develop/docker
It would be nice to have the docker file setup the database and everything as well so that it just shot you straight to configure_system.php maybe with some values pre-initalized.

AricStewart:
This is of course presuming that the docker course of action is around setting up a development or demonstration setup.  If we where going to target a production docker container we would need a way to configure the passwords and such.

mshappe:
That's exactly the kind of thing I had in mind for the next iteration -- something that gets everything ready for `configure_system.php`. This is not QUITE there yet, but yeah, that's what I had in mind for the next phase.

I deliberately kept this file named `Dockerfile.dev` to make room for a more production-oriented Dockerfile like the one we've discussed in Slack.

AricStewart:
Fair.. Fair..
2nd question... Do we want to put this into a subdir such as docker/ or something so that if/when the files involved grow they do not complicate the root of the project (which already has a lot of files) Or would that involve too much complication?

mshappe:
An capital idea! Probably not tonight, but I will iterate on this with that excellent example above in mind in the next day or so.

mshappe:
So, I've implemented a trimmed down version of the entrypoint idea (with room for later expansion but right now this is mainly for development) -- most importantly it will now run composer on its own. If you run it with `docker-compose up` it will mount the source tree over what it copies into the Docker image, which also means when it populates `vendor`, you'll see it in your local as well. 

I didn't wind up putting them in their own separate directory because there are some context oddities when you do that, and I hadn't quite worked them all out (I've never actually separated them out like that before in other projects). I will come back to that at a later date, I think.

The last bit I haven't added yet is a DOCKER-README.md or something, which I should do.

AricStewart:
That is looking really good!  I would still like the readme or something.
My goal is that me, as a competent developer but only a few time user of docker, should be able to get this up and running in docker with limited pain. 

So starting the DOCKER-README.md with "Install and configure docker on your platform..." For example is just fine.  No need to target the lowest rung at this point. 

How hard would it be to get the readme?  Then is it in a point where I could get it running on my machine with docker pretty easily? 

I am excited for this.

mshappe:
It's easy. I just ran out of brain last night :-)

AricStewart:
Understandable. Mind if I wait for it so I can test it before approving the pull request?

AricStewart:
Docker is now in the project.

Issue #253, opened 2019-04-09
Registration Worker: Guest Lookup


seanbseanb:
Registration Workers should be able to look up guest registrations by last name, email address , or badge number.

seanbseanb:
I believe last name/first name is already closed, pending review.

AricStewart:
Any of the lookup bars should respond to full names (Real or preferred) , email and badge number

https://ciab.anergyst.org/index.php?Function=registration

If you need to be able to look someone up by JUST last name I can add that facility.

AricStewart:
Please test the functionality.

AricStewart:
Check it out again please.. updated:
https://ciab.anergyst.org/index.php?Function=registration&memberId=1231

mshappe:
What is the current status of this issue?

AricStewart:
I believe this is complete and just was waiting on official testing/closing.

AricStewart:
I am going to poke this one last time and then just close it as fixed and inactive.

mshappe:
This appears to work as intended.

PR #254, opened 2019-04-09
Allow partion user matching for broader lookup

AricStewart:
This patch will need to be rewritten to move forward off Neon
Test this functionality currently at:
https://ciab.anergyst.org
Under the Member Registration tab

### DIFF
--- a/functions/users.inc
+++ b/functions/users.inc
@@ -99,73 +99,199 @@ function lookup_user_by_id($id)
 }
 
 
-function lookup_users_by_email($email, $single = true)
+function lookup_users_by_email($email, $single = true, $substring = false)
 {
-    return _lookup_user([['Email', 'EQUAL', $email]], $single);
-
-}
-
-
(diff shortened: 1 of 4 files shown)


AricStewart:
For Issue #253

functions/users.inc
```diff
@@ -99,73 +99,185 @@ function lookup_user_by_id($id)
...
-    /* We have to do a few searchs due to Neon stupidity */
-    $output = _lookup_user(
-        [[$preferredFirstName, 'EQUAL', $names[0]],
-         [$preferredLastName, 'EQUAL', $names[1]]],
-        $single
-    );
-    if (count($output['users']) != 0) {
-        return $output;
+    if ($result2['code'] == null) {
+        if ($output['code'] === null) {
```
mshappe: Why `===` here, but `==` above? I know they're subtly difference; just not clear if the difference matters here.

PR #255, opened 2019-04-11
Pending badge printing

AricStewart:
A lot has changed here but most of it is the fault of the style changes i was doing.

I believe this should be full functional and ready for testing.


functions/functions.inc
```diff
@@ -182,3 +182,30 @@ SQL;
...
+    header("Last-Modified: ".gmdate("D, d M Y H:i:s")." GMT");
+    header("Cache-Control: no-store, no-cache, must-revalidate");
+    header("Cache-Control: post-check=0, pre-check=0", false);
+    header('Content-Disposition: attachment; filename="'.$filename.'"');
+    header("Pragma: no-cache");
+
+}
+
+
+function output_cvs($report, $filename)
```
mshappe: `output_csv` not `cvs`. Unless we can suddenly generate drug stores :-D

Issue #261, opened 2019-04-12
Supportability: Understanding deployed versions


seanbseanb:
When looking at the front page of CIAB, there should be a version number findable within either the source, or in an unobtrusive place on the screen.  In other products, this is sometimes shown with a full “Version: 1.2.3.4” tag in clear text, in other products it has been either hidden in the page source, or shown as white-on-white text (which can then be shown by “select-all” on the main page.

AricStewart:
This issue is now resolved.

PR #262, opened 2019-04-12
More detailed versions string

AricStewart:
For Issue #261

### DIFF
--- a/functions/functions.inc
+++ b/functions/functions.inc
@@ -6,8 +6,6 @@
 
 // Main setup function
 $MODULENAME  = "Member Portal";
-$BUILD       = "154";
-$VERSION     = "1.3 - Captain America, Build: ".$BUILD; // Major.Minor.Revision Format - This is a string, version names allowed
 
 // Discover Self Location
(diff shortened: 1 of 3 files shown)


functions/functions.inc
```diff
@@ -25,6 +23,52 @@ $MODULESDIR     = $BASEDIR."/modules";
 // Set a Date stamp for use anywhere
 $DS             = uniqid('', true);
 
+
+function getVersion()
```
mshappe: Hm. I think I'm OK with this for now, but I dislike having a requirement that `git` exist on the target server. The way we're deploying now, it's a safe assumption, but if we were to move to a very different model (say, docker), it might not be.

AricStewart: Understood. The idea is that a Release version without git would have "build" set properly in the verison.ini file. 

I am toying with writing some sort of a packager that can be run to generate a nice .zip or .tgz file that is ready for deployment.  Does all the composer stuff and updates the "build" property etc... But that may not be worth it at this point.

Issue #190, opened 2018-07-18
Emailer Interface module


tskeeley:
Need the ability to send emails from the system, to:
 - All
 - All registered for one/many events (if reg module)
 - All volunteers for one/many events (if vol module)
- All ConCom (if concom module)
 - All Dealers/Invited Participants (based on reg module)
 - All Artists (if artshow module)
 - All winning Art Bids (if artshow module)
 - Upload-able email list

Must support "opt-out", "do-not-contact", "Deceased", and Opt-In (GRPD)

Opt-out/contact info is in NeonCRM, need to capture.

Should support common lists (above), along with a cut/paste way of entering the html and plain-text way to send (might not need plain text?)  Should we support a simple WYSIWYG editor (good open source ones that would could dump in without issue)

AricStewart:
A few questions my head has come up while designing this, since that is what my brain does

- Do you want to be sending e-mails from the web site?
  - Dropdown From selector?
  - Always from admin?
  - Allow a reply-to? How would it be selected/determined?  
- Who has access to this tool?
- Where is it located?
  - Is it located under 'Member Registration' or maybe we change that to 'Membership' and we need sub-panels for 'Registration' etc.. or something. 
  - Would it maybe be located in each module? 

I did research for a good WYSIWYG editor before for a work project and found one. I will look to see if I can find it again.

tskeeley:
Good questions!  And good evidence that I completely failed to notice you ased about this!

Emails from the website, via wysiwyg would be the best.  If wysiwyg editing is outside of scope, then we should look at linking into "Constant Contact" or " Vertical Response" (one of those has a VERY good non-profit deal available, as long as we can API address data).  

From selector - Yes, dropdown, limited to pre-approved departments (editable by admin)

From admin - No, department heads could send emails IF* they are so approved to do so.  I think this is a setting.  In the case of CVG, it's likely that this permission would go to the head of publications or social media, but it might be limited to whatever director is in charge of communications.

reply-to - selectable, like from: field.  (there are times when a mail might go out but replies should go to "no-reply" or "info@"

location - I'm tempted to say that this is a core function, at least based on registration.  But, it's not something anyone but admins and approved senders would ever see.  Seems like something that would "show" in the admin menus but come from the registration/membership module.  I"m open to differing opinions on that.

AricStewart:
This framework is in place. smaller bugs should be opened for specific changes in functionality as required/requested.

PR #265, opened 2019-04-29
Emailer Module

AricStewart:
For Issue #190 
Fully Functional, I believe, new e-mail lists can only be created by an admin. But once given proper access others can edit the lists and send e-mail to the lists.

### DIFF
--- a/modules/emailer/functions/emailer.inc
+++ b/modules/emailer/functions/emailer.inc
@@ -1,0 +1,472 @@
+<?php namespace mailer;
+
+/*.
+    require_module 'standard';
+    require_module 'filter';
+.*/
+
+require_once $FUNCTIONDIR."/divisional.inc";
+require_once $FUNCTIONDIR."/database.inc";
+
+class Email
+{
+
+    private $user;
+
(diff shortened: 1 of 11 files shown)


modules/emailer/functions/emailer.inc
```diff
@@ -0,0 +1,472 @@
...
+    public function send($data)
+    {
+        $to = $this->getToById($data->to);
+        $headers = 'From: '.$data->from."\r\n";
+        $headers .= 'Reply-To: '.$data->reply;
+        $headers .= "MIME-Version: 1.0\r\n";
+        $headers .= "Content-Type: text/html; charset=ISO-8859-1\r\n";
+        if (count($to) > 0) {
+            foreach ($to as $target) {
+                mail($target, $data->subject, $data->body, $headers);
```
mshappe: So: since PHP can be configured so that `mail` uses SMTP, we should not have to change **anything** if we switch to using Mailgun or something like it with their SMTP-based interface--just make sure our PHP environment is configured to use it.

PR #268, opened 2019-07-05
Do not add 12 to 12PM when converting to 24hour time

AricStewart:


mshappe:
Who can actually merge? @tskeeley @AricStewart

PR #271, opened 2019-07-10
When generating department reports only generate for current year

AricStewart:


functions/volunteer.inc
```diff
@@ -602,8 +602,12 @@ SQL;
 }
 
 
-function volunteers_in_dept($department)
+function volunteers_in_dept($department, $event = null)
 {
+    if ($event === null) {
+        $event = current_eventID();
```
mshappe: In PHP, can the initializer in the argument list be a method call? That is, can you just say

```php
function volunteers_in_dept($department, $event = current_eventID())
```

and skip a step?

modules/volunteers/admin/pages/pre.inc
```diff
@@ -44,8 +44,9 @@ if (!empty($_REQUEST)) {
         if (isset($_SESSION['IS_ADMIN']) || $_SESSION['IS_VOLUNTEERS']) {
             $dept = $updateData['dept_report'];
             $name = $updateData['dept_name'];
-            $report = volunteers_in_dept(intval($dept));
-            output_csv($report, "department_report_".$name.".csv");
+            $event = current_eventID();
```
mshappe: Redundant with the way you're already initializing this in `volunteers_in _dept`.

Also, what if I really **want** a list of vols across all years?

functions/volunteer.inc
AricStewart: I do not believe so. Or at least a few minutes of web searching did not turn up a way to do that. And quick test programs do not work.

modules/volunteers/admin/pages/pre.inc
AricStewart: Yes I am being redundant but part of me is waiting for the request to run reports or do actions on historical data. So as long as I keep things able to do that it makes it easier to fufill that potential future request.

AricStewart: as for the multiple years that makes sense.  I will work it out to allow for that.

AricStewart: Patch updated to allow for multiple years or all years. So when/if we add UI for that we can make it work.

functions/volunteer.inc
mshappe: 👍

modules/volunteers/admin/pages/pre.inc
mshappe: 👍

mshappe:
I'm still not authorized to merge, so that'll have to be @tskeeley

AricStewart:
> I'm still not authorized to merge, so that'll have to be @tskeeley

You should be able to now. try again.

seanbseanb:
Can I drop this one into prod, or will it conflict with other commits?

PR #274, opened 2019-07-16
Allow all of concom to upload hours

AricStewart:
Addresses issues Programming had and have been discussed in e-mail

### DIFF
--- a/modules/volunteers/enter/upload/pages/pre.inc
+++ b/modules/volunteers/enter/upload/pages/pre.inc
@@ -5,8 +5,12 @@
     require_module 'json';
 .*/
 
-/* This should only be accessable from volunteer staff */
-if (!(isset($_SESSION['IS_ADMIN']) || $_SESSION['IS_VOLUNTEERS'])) {
+/* This should only be accessable to CONCOM */
+$concom = null;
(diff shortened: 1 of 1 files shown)


modules/volunteers/enter/upload/pages/pre.inc
```diff
@@ -5,8 +5,12 @@
     require_module 'json';
 .*/
 
-/* This should only be accessable from volunteer staff */
-if (!(isset($_SESSION['IS_ADMIN']) || $_SESSION['IS_VOLUNTEERS'])) {
+/* This should only be accessable to CONCOM */
+$concom = null;
+if (class_exists('\\concom\\POSITION') && method_exists('\\concom\\POSITION', 'getConComPosition')) {
```
mshappe: Is there a reason why the class would not exist?

AricStewart: Modules in the 'modules' directory are designed to to be able to be explicitly disabled. It is a rough plugin-in framework to try to help get things that could be optional to other conventions.

So that is just making sure the concom module has been loaded and the method we are wanting exists in that module.

mshappe: Makes sense. Fire when ready.

PR #276, opened 2019-07-20
Aric improve csv import

AricStewart:
Cache user lookup for speed and reformat date

### DIFF
--- a/modules/volunteers/enter/upload/pages/pre.inc
+++ b/modules/volunteers/enter/upload/pages/pre.inc
@@ -18,6 +18,23 @@ require_once($FUNCTIONDIR."/volunteer.inc");
 
 $response = array();
 
+$user_cache = array();
+
+
+function lookup_user($user)
+{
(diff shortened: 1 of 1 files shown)


modules/volunteers/enter/upload/pages/pre.inc
```diff
@@ -43,28 +44,55 @@ while (!feof($file)) {
         }
         /* user lookups */
         $user = null;
-        $output = lookup_users_by_key($csv[0]);
+        if (array_key_exists($csv[0], $user_cache)) {
```
mshappe: Why not make the cache lookup part of `lookup_users_by_key`? One way or another, this repeated idiom seems like it should be DRY'd up...

AricStewart: Ok updated, I did not go the route of having the case in `lookup_users_by_key` but I did pull it out into its own function to have less code repetition and clean things up.

PR #277, opened 2019-07-26
Lots of scss work

AricStewart:
This completely removes direct w3- styles from all of the base pages and functions.


functions/neon.inc
```diff
@@ -257,109 +257,6 @@ class Neon
   public function rebuildNeonData() {
     ##==## Boy is this gonna be fun!
   }
-  
-  public function doGeneralSearch() {
```
mshappe: Why is this going away?

AricStewart: As far as I can tell it is completely unused and so since we are on the path to removing neon though to just remove this function instead of fixing the w3- class tags used with in

PR #280, opened 2019-08-11
Do not count promo items in spent hours

AricStewart:


functions/volunteer.inc
```diff
@@ -112,6 +112,14 @@ function get_volunteer_year_summery($event = null)
...
                 WHERE
                     EventID = $event
+                    AND (
+                        SELECT
+                            Promo
+                        FROM
+                            VolunteerRewards v
+                        WHERE
+                            v.prizeID = h.prizeID
+                       ) != 1
```
mshappe: I take it `Promo` is a boolean flag, here?

Issue #292, opened 2019-10-04
Division structure: add department button broken


mshappe:
The little yellow plus sign spouts a javascript error in the console instead of doing what's needed to allow adding a department.

AricStewart:
Should now be fixed.

mshappe:
Possibly, but it's not yet deployed anywhere I can see it :-D

AricStewart:
True, I have been closing bugs when things have been merged into the master branch, mostly because for a long time deliveries just where never happening.

Do we want to change that process?

mshappe:
Lemme think about that. For now, continue as you have been. I guess the main point is that I need to stick Thomas with a pointy thing to get things deployed since they're still not autodeploying.

PR #294, opened 2019-10-04
Fix Adding New Departments and Divisions to ConCom Structure

AricStewart:


AricStewart:
This address issue #292

AricStewart:
It was working, then functionality was expanded and was not re-tested. It is working again. :)

PR #295, opened 2019-10-04
Add a framework to allow for configurable backend email services

AricStewart:


backends/email.inc
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+
+    public static function instance()
+    {
+        if (self::$instance === null) {
+            $client = $_ENV['EMAIL_BACKEND'];
+            if (!$client || empty($client)) {
+                $client = 'mail.inc';
+            }
```
mshappe: Clever way to handle it and retain current default behaviour!

PR #298, opened 2019-10-09
Start a Role Based Access Control structure

AricStewart:
This will allow the distribution of finer grain access to various parts of the site based on ConCom position.


data/database-schema.php
```diff
@@ -153,6 +153,13 @@ class SCHEMA
         'EmailListID' => 'INT UNSIGNED NOT NULL',
         'EditList' => 'BOOLEAN NOT NULL',
         'ChangeAccess' => 'BOOLEAN NOT NULL',
+    ],
+
+    'ConComPermissions' => [
```
mshappe: You'll want to add an index on `Permission` -- we'll be searching on it a lot, I suspect.

AricStewart: Ok change made to support that and pushed.

Issue #312, opened 2019-10-20
Portal: Concom Structure ordering


mshappe:
It appears that departments are showing up in the order they've been added (newest ones last within a division) rather than alphabetically. They should either be automatically alphabetized, or re-orderable by dragon-drop. Don't really care which :-)

PR #313, opened 2019-10-20
Clean up Departemnt ordering

AricStewart:
For Issue #312

### DIFF
--- a/functions/divisional.inc
+++ b/functions/divisional.inc
@@ -68,7 +68,7 @@ SQL;
             $value = $result->fetch();
         }
     }
-    asort($Departments);
+    ksort($Departments);
     asort($Divisions);
 
     // Push non-divisions to the end of the list (force order)


functions/divisional.inc
```diff
@@ -68,11 +68,12 @@ SQL;
...
     }
-    asort($Departments);
-    asort($Divisions);
 
     // Push non-divisions to the end of the list (force order)
     $endDivs = ['Board Staff', 'Committees', 'Corporate Staff'];
     $Divisions = array_diff($Divisions, $endDivs);
     $Divisions = array_merge($Divisions, $endDivs);
+
+    asort($Divisions);
```
mshappe: Should this one still come before the bit where we push `$endDivs` to the end? It's the departments within a division I was concerned about.

AricStewart: Ok change pushed. Re-review please

PR #314, opened 2019-10-21
Add sendgrid email backend.

AricStewart:
To enable this backend set the following in .env or the environment

EMAIL_BACKEND="sendgrid.inc"
SENDGRID_MAIL_FROM="<email address>"
SENDGRID_API_KEY="<apikey>"

### DIFF
--- a/backends/email/sendgrid.inc
+++ b/backends/email/sendgrid.inc
@@ -1,0 +1,64 @@
+<?php
+/*
+ *  This is a email back end module using the Official Twilio SendGrid
+ *  Led, Community Driven PHP API Library
+ *  https://sendgrid.com
+ */
+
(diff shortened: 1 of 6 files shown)


backends/email/sendgrid.inc
```diff
@@ -0,0 +1,64 @@
...
+        $message,
+        $additional_headers = null,
+        $additional_parameters = null,
+        $from = null
+    ) {
+        if ($from == null) {
+            $from = getenv('SENDGRID_MAIL_FROM');
+        }
+        $key = getenv('SENDGRID_API_KEY');
+        if ($form == null || $key == null) {
```
mshappe: `$form` here should probably be `$from`

PR #316, opened 2019-10-22
Aric improve email

AricStewart:


functions/functions.inc
```diff
@@ -262,3 +259,19 @@ function output_csv($report, $filename)
     }
 
 }
+
+function urlsafe_b64decode($string) {
```
mshappe: You define these here, and then again, very differently, in common...

AricStewart: One is PHP and one is Javascript.

AricStewart: One is PHP and one is Javascript.

mshappe: ...derp *facepalm*

PR #337, opened 2019-11-06
Initial docker starting point

AricStewart:
This can of course be expanded and refined as we move forward. But it gives a baseline.


mshappe:
@jason-edstrom Please check this, as well!

PR #353, opened 2019-11-09
Aric neon sidebyside

AricStewart:


AricStewart:
Ok rebased back onto the master tip with your docker changes.

mshappe:
Worst that happens is it explodes!

PR #369, opened 2019-11-15
Make some docker tweaks

mshappe:
These are not actually side-by-side-mode tweaks per se, but they've been tested mainly on that branch.


AricStewart:
I will play some games and get these merged into master.

AricStewart:
Ok merged into master!

PR #374, opened 2019-11-16
Update to font awesome 5

AricStewart:


AricStewart:
> I read that the `far` prefix is only available with a pro license. Do we have that?

Strange `far` is working just fine. They are some part of base and some not. We only use ones that are part of base...

PR #387, opened 2019-11-25
Neon webhooks

AricStewart:


AricStewart:
Sure.  So NeonCRM provides a limited way to receive POST data when something in the neon database changes.

https://developer.neoncrm.com/api/webhooks/

This will mean if people add memeberships or change data via NEON then we can cleanly update that data in CIAB behind the scenes without them having to log into CIAB. 

The goal here is to make the CIAB database as complete and up to date as possible without having to scrape the Neon database on some sort of a cronjob. 

Sadly webhooks are limited and do not include, say, ticket data...

mshappe:
Oh, I get that part (although other people reading this might not have known that part :-D ). The problem is that I don't see for myself exactly how this works to make that happen. I get a vague sense that some of this is configuration and some of it actually handles the inbound POST, but that's about all.

mshappe:
(NOTE: this is more because the structure of this app is still very different from most of what I work on these days, and not because you're not writing clear code.)

AricStewart:
No worries.  So when you add a webhook it adds the given URL and a randomly generated key to the Neon WebHook API registering for the CREATE_ACCOUNT and UPDATE_ACCOUNT events.

We have all the code to import a user from Neon in place right now. So when that webhook comes in we can actually ignore all the data there and instead just do a quick search for that user, which will auto-import  and update that neon data in our local database. 

This is just a stepping stone to  keep our data consistant until we turn off the Neon piece all together.

PR #388, opened 2019-11-25
Staff and position wide RBAC permissions

AricStewart:


modules/concom/functions/RBAC.inc
```diff
@@ -64,6 +71,10 @@ class ConComRBAC
                 foreach ($positions as $pos) {
                     $new = $value['DepartmentID'].'.'.$pos;
                     $rbac->rbac->addRole($new, $parent);
+                    $role = $rbac->rbac->getRole($new);
+                    $staff->addParent($role);
+                    $target = $rbac->rbac->getRole("all.$pos");
+                    $target->addParent($role);
```
mshappe: This code should get DRY'd up. The entire loop looks like it's repeated verbatim at least twice.

AricStewart: Good catch. Code duplication is definitely one area where I am not very rigorous.

PR #390, opened 2019-11-25
Split Volunteer reports to separate report menu

AricStewart:
Groundwork for further report generation.

### DIFF
--- a/modules/volunteers/report/pages/body.inc
+++ b/modules/volunteers/report/pages/body.inc
@@ -1,0 +1,181 @@
+<?php
+
+
+function _htmlsafeify($array)
+{
+    $output = [];
+    foreach ($array as $key => $value) {
+        $output[$key] = htmlspecialchars($value);
+    }
+    return $output;
+
+}
+
+
+function hour_table()
+{
+    $concom = null;
+    $data = get_volunteer_hour_summary();
+    $summary = get_volunteer_year_summery();
+    if (!empty($data)) {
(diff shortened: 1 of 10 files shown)


modules/volunteers/report/pages/body.inc
```diff
@@ -0,0 +1,181 @@
...
+
+  <div class='UI-sidebar-hidden UI-fixed' id='csv_export_derived_div'>
+    <div class='UI-center'>
+      <h2 class='UI-red'>Export Derived CSV Data</h2>
+    </div>
+    <div class='UI-center'>
+      <label class='UI-label' for='der_csv_table'>
+      Table:</label>
+      <select class="UI-margin UI-select" style="width:auto" id="der_csv_table">
+        <option value="SELECT r.*, (SELECT COUNT(h.PrizeID) FROM HourRedemptions AS h WHERE h.PrizeID = r.PrizeID GROUP BY h.PrizeID) as Claimed FROM VolunteerRewards as r;">Volunteer Rewards</option>
```
mshappe: Um...let's NOT include SQL as possible values here. We do not want to get into the habit of accepting and executing SQL submitted by a form, ever.

AricStewart: Ah yes... that is pre-existing for the record. Still horrible and needs to be fixed but pre-existing. :)

AricStewart: addressed

PR #392, opened 2019-11-26
More request backends

AricStewart:


sitesupport/common.js
```diff
@@ -53,16 +53,20 @@ function urlsafeB64Decode(data) {
...
   xhttp.onreadystatechange = function() {
     if (this.readyState == 4 && this.status == 200) {
-      finish(this);
+      success(this);
       hideSpinner();
-    } else if (this.status == 404) {
-      alertbox('404!');
-    } else if (this.status == 409) {
-      alertbox('409!');
+    } else if (this.readyState == 4) {
```
mshappe: Out of curiosity, what exactly does `4` mean for `readyState`?

Perhaps, more broadly, we should introduce some constants here to make this self-documenting :-D

PR #395, opened 2019-11-28
Work on site settings management

AricStewart:


backends/email.inc
```diff
@@ -28,7 +28,7 @@ class Email
     public static function instance()
     {
         if (self::$instance === null) {
-            $client = $_ENV['EMAIL_BACKEND'];
+            $client = $GLOBAL['EMAIL_BACKEND'];
```
mshappe: `$GLOBAL` or `$GLOBALS`?

AricStewart: doh! good catch!

Issue #307, opened 2019-10-15
Portal: Concom List: A person's mambership should be editable, not just remove/add-again


mshappe:
It should be possible to edit (and hence correct) a person's entry on the concom list. For example, I should be able to select Lauren under Member Advocates and add/change a note to refer to the fact that she's actually a Board Liaison.

(Interestingly, removing her and re-adding her, with the note, didn't work -- she showed up without the note, still)

AricStewart:
Fixed the note part but I want to know some of the work flow here.

- Changing Notes:  Easy
- Changing Positions in a department.  Can only be done by department head (positions < head) or  Divisional member
- Changing Departments, If in the same Division then can be done by a Division member. If between 2 Division only Admin? 

What about movement within Divisions. Only doable by Admins?

thanks.

mshappe:
The only things I think need to be editable (not add/drop) right now are Notes and Position-within-department (both restricted to Department Head and above). Changing departments can be drop/add; changing divisions can be drop/add.

mshappe:
Had occasion to test this just yesterday for one of my departments, so we're good!

Issue #302, opened 2019-10-15
Portal: Department Management: "removing" a member should just flag their assignment as "ended"


mshappe:
So that we can keep track of historic department membership, rather than actually deleting a membership record when someone is removed from a department, their membership should be flagged with an `EndedOn` datestamp. The main concom list could then just display those for whom `EndedOn IS NULL` and we can build out a way of looking at historic membership later.

AricStewart:
I believe this is in and working. Please test.

PR #403, opened 2019-12-10
concom list work

AricStewart:
This pass is mostly cleaning up and modernizing.

A little work with the flow of adding members.

Groundwork for issues #307 and #302

### DIFF
--- a/modules/concom/functions/concom.inc
+++ b/modules/concom/functions/concom.inc
@@ -49,238 +49,347 @@ function getPositionID($position)
 }
 
 
-function printTableHeader($section, $long = true)
+function conComRBAC($instance)
 {
-    echo "<tr id=\"table_head_$section\" class=\"CONCOM-table-head\">";
-    if ($long) {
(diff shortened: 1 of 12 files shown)


AricStewart:
Note that this does change the look and feel of the concom list a fair bit. Please review that as well please.

modules/concom/functions/concom.inc
```diff
@@ -49,236 +49,344 @@ function getPositionID($position)
...
-    foreach ($list as $line) {
-        $printList[$line['Division']][$line['Department']][$line['Position']][$line['First Name']][$line['Last Name']] = $line['Email'].'|'.$line['Note'].'|'.$line['Account ID'];
+}
+
+
+function printDivisionHeader($div)
+{
+    global $Divisions;
+
+    echo '<div id="'.$div.'" class="UI-table UI-table-heading">';
```
mshappe: This is where I start longing for templates...

PR #411, opened 2020-02-12
The beginning of an API and Deadlines

AricStewart:


api/src/Controller/Department/GetDepartment.php
```diff
@@ -0,0 +1,82 @@
...
+                $email[] = [
+                'email' => $entry['EMail'],
+                'isAlias' => $alias
+                ];
+            }
+            $output['name'] = $output['Name'];
+            $output['division'] = $output['Division'];
+            $output['fallback'] = $output['FallbackID'];
+            $output['email'] = $email;
+            unset($output['Email']);
```
mshappe: Is there a way in PHP to dry this up? In Ruby, for example, I could say something like:
```ruby
%w(Email Name Division Fallback FallbackID).each { |key| output[key] = nil }
```

AricStewart: Part of the trick is there is not clean 1:1 there.  Capitalization is different and a few have different keys (fallback vs FallbackID)

mshappe: I meant mainly the list of unsets.

AricStewart: Ah the part UNDER the comment as opposed to ABOVE the comment. ;)
I do not know if there is a clean PHP way to dry that up.  All the PHP documentation online tell you to do the assignment and an unset. 
Though we could get tricky with array_map but it starts to become much less clear code in my opinion.

PR #418, opened 2020-03-07
Adjust Neon sync to handle foreign keys

AricStewart:
This is used by Announcements and was causing neon updates to fail for accounts that have made an announcement


mshappe:
I may not be at a real keyboard at all today, but I have looked this over
and it's approved, so go ahead when you're ready. Interesting that REPLACE
is so different from INSERT...ON DUPLICATE KEY UPDATE.

PR #453, opened 2020-11-06
Aric/new lost password

AricStewart:
[#175217363]


AricStewart:
This patch is up and testable at:

https://ciab.anergyst.org/

api/src/App/Routes.php
```diff
@@ -6,6 +6,7 @@
 $app->get('/member/{name}/status', 'App\Controller\Member\GetStatus');
 $app->post('/member[/]', 'App\Controller\Member\PostMember');
 $app->post('/member/{name}/password', 'App\Controller\Member\PostPassword');
+$app->put('/member/{name}/password/recovery', 'App\Controller\Member\PutPassword');
```
mshappe, line 9: All of these should also change eventually to be by ID. But that can be done separately.

mshappe (approved):

PR #462, opened 2020-12-19
Spike stripe

mshappe:
* Introduce the MVP backend pieces needed to redirect to a Stripe checkout page and have it redirect back to us on success
* Introduce the dead minimum handler for Stripe webhooks (does nothing but logs them right now)
* Introduce a "Registration Store" module which right now just has a dumb Checkout button that buys an angry ferret
* Add the success page that the Stripe Hosted Checkout will redirect to when you successfully buy a ferret.

This code, as is, is never meant to merge to Master without a lot more stuff, and is PR'd here only for review of approach.


AricStewart:
Doing a closer review a few comments (since you are working on this)

We are trying to avoid tables, tables suck for responsive displays.
We are using <div>s and  the "UI-table..." classes.  That should give us css tables.  THey are less powerful than old fashion table but far more responsive.

```
            <div id="my_table_id" class="UI-table-all">
                  <div class="UI-table-header">
                    <div class="UI-table-row">
                      <div class="UI-table-cell">
                            header1
                      </div>
                     <div class="UI-table-cell">
                            header2
                      </div>
                    </div>
                  </div>
                  <div class="UI-table-body">
                    <div class="UI-table-row">
                      <div class="UI-table-cell">
                            First Item
                      </div>
                     <div class="UI-table-cell">
                            Second Item
                      </div>
                    </div>
                  </div>
            </div>
```
You can look in `scss/common.scss` for those styles

mshappe:
I remember us having this conversation before.

Tables as an alternative to genuine layout are bad.

Tables for actual tabular data are fine AND can be made adequately
responsive.

AricStewart:
Ah, huh.  Well my html foo is very weak. I just had it in my brain that tables == bad.  :) I will trust you. 

Also your stuff around scss_cache/scss may not be needed anymore.  I hope I fixed that issue. If not could you confirm and I will attack it again?

AricStewart:
Inactive and likely will need rework to integrate.

PR #463, opened 2021-01-03
Aric/member check in

AricStewart:
This completes the entire member check-in/badge pickup story.

### DIFF
--- a/api/src/Modules/registration/Controller/Ticket/BaseTicket.php
+++ b/api/src/Modules/registration/Controller/Ticket/BaseTicket.php
@@ -1,0 +1,155 @@
+<?php declare(strict_types=1);
+/*.
+    require_module 'standard';
+.*/
+
+namespace App\Modules\registration\Controller\Ticket;
+
+use Slim\Container;
+use Slim\Http\Request;
+use Slim\Http\Response;
+use App\Modules\registration\Controller\BaseRegistration;
+
+abstract class BaseTicket extends BaseRegistration
+{
+
+    public $id = 0;
+
+
+    public function __construct(Container $container)
+    {
+        parent::__construct($container);
+
+    }
+
+
+    protected function buildTicketHateoas(Request $request, $id)
+    {
+        if ($this->id !== 0) {
+            $path = $request->getUri()->getBaseUrl();
+            $this->addHateoasLink('self', $path.'/registration/ticket/'.strval($id), 'GET');
+            $this->addHateoasLink('checkin', $path.'/registration/ticket/'.strval($id).'/checkin', 'PUT');
+            $this->addHateoasLink('lost', $path.'/registration/ticket/'.strval($id).'/lost', 'PUT');
+            $this->addHateoasLink('pickup', $path.'/registration/ticket/'.strval($id).'/pickup', 'PUT');
+            $this->addHateoasLink('email', $path.'/registration/ticket/'.strval($id).'/email', 'PUT');
+            $this->addHateoasLink('print', $path.'/registration/ticket/'.strval($id).'/print', 'PUT');
+        }
+
+    }
+
+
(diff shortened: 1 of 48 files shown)


api/src/Modules/registration/Controller/Ticket/BaseTicket.php
```diff
@@ -0,0 +1,155 @@
...
+            $this->addHateoasLink('lost', $path.'/registration/ticket/'.strval($id).'/lost', 'PUT');
+            $this->addHateoasLink('pickup', $path.'/registration/ticket/'.strval($id).'/pickup', 'PUT');
+            $this->addHateoasLink('email', $path.'/registration/ticket/'.strval($id).'/email', 'PUT');
+            $this->addHateoasLink('print', $path.'/registration/ticket/'.strval($id).'/print', 'PUT');
+        }
+
+    }
+
+
+    public function buildTicket($base, $ticket)
```
mshappe, line 41: This idiom is one of the few things that makes me wish we were using an ORM with model classes. Being able to put transformations on the model would be nice. Otherwise, I mostly don't miss it.

api/src/Modules/registration/Controller/Ticket/PostTicket.php
```diff
@@ -0,0 +1,96 @@
+<?php declare(strict_types=1);
+/*.
+    require_module 'standard';
+.*/
+
+namespace App\Modules\registration\Controller\Ticket;
```
mshappe, line 6: Is there a reason `registration` is in lowercase, throughout? We've been using sentence case I thought, mostly.

modules/registration/import.js
```diff
@@ -0,0 +1,92 @@
...
+      if (data.length > 0) {
+        var eventName =  data[0].event.name;
+        var generated = 0
+        data.forEach(function(ticket) {
+          if (ticket.BoardingPassGenerated) {
+            generated += 1;
+          }
+        });
+        if (generated < data.length) {
+          apiRequest('GET', 'registration/open', null)
```
mshappe, line 42: Once we get into really long promise chains, I like to actually pull the functions up and name them, then it can be

```javascript
foo()
  .then(bar())
  .then(baz())
  .then(quux())
  .catch(everythingIsFucked());
```

mshappe (commented):
Awesomesauce. Just a few comments/questions!

api/src/Modules/registration/Controller/Ticket/BaseTicket.php
AricStewart: Sorry missed these comments.  Yeah I have not used many ORM model classes so I am unfamiliar with the benefits and difficulties.

api/src/Modules/registration/Controller/Ticket/PostTicket.php
AricStewart: registration is lower case because the module name need to match the module name in /modules/ in the base of the source.  This connects the API implementation with the feature module.  The php modules in /modules are all lowercase so that ended up having this lower case.  Now that could be fixed with a bit of case insensitivity or selective case insensitivity if we wanted to add that complexity.

modules/registration/import.js
AricStewart: Most of the long promise chains I am runngin into are more
foo()
 .then( function() {
    .then( function() {
     })
 })

So nested instead of sequential.  Unless I am misunderstanding it we end up needing to next them like that.

mshappe:
There is no need to nest. Just have the function return the next promise
and the chained "then" attaches to that. But even if there were, raising
the functions upscope and naming them still cleans up a lot of hard to read
depth and makes the code self-documenting.

I'm typing with thumbs right now but I'll try to find a good exemplar for
you later today.

PR #470, opened 2021-01-24
Improve table creation to handle columns correctly 

mshappe:
…and allow for multi-column indexes correctly

### DIFF
--- a/data/database-schema.php
+++ b/data/database-schema.php
@@ -150,5 +150,11 @@ class SCHEMA
 
     ];
 
-    public static $DB_index = [];
+    public static $DB_index = [
+        'Members' => [
+            'idx_Email' => ['Email'],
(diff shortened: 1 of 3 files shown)


tskeeley (approved):
So, moving to a mutli field index.  fair enough! :)  That seems to make sense to me.

I also like that you added the errorcode check.  That makes me happy

mshappe:
Yes and no.

Each one of those hash entries will create its own index on a specific column.

If you specify more than one column in an entry, however, like if I added one that was `Name => [FirstName, LastName]`, THAT would produce a compound index.

What you had before would have ALWAYS produced a single compound index if given a list of columns, which I'm pretty sure was not your intent!

AricStewart (approved):

PR #471, opened 2021-01-24
Block duplicated email

AricStewart:
This transition is going to be a bit timing sensitive.

You will want to run the clean_duplicate_accounts tool via the web
when there is no neon update running.  Have ?update=1 to do the work
otherwise it will be a dry-run.

Then when the next neon update occures it should properly
not re-duplicate accounts and new neon accounts should not end up as
duplicates.

However if a neon update occures after this patch is applies and
before you run clean_duplicate_accounts, then the neon update will
behave a bit unpredictable around accounts with duplicate email and
may not assige the correct primary account.

### DIFF
--- a/tools/clean_duplicate_accounts.php
+++ b/tools/clean_duplicate_accounts.php
@@ -1,0 +1,134 @@
+<?php
+/*.
+    require_module 'standard';
+.*/
+
+require_once(__DIR__."/../functions/functions.inc");
+
+$dependsField = 'dependentOnID';
+$do_update = false;
+
+
+function find_duplicates()
+{
+    $sql = <<<SQL
+        SELECT
+            *
+        FROM
+            `Members` AS base
+        WHERE
+            (
+            SELECT
+                COUNT(AccountID) AS COUNT
+            FROM
+                `Members`
+            WHERE
+                Email = base.Email
+        ) > 1
+        ORDER BY
+            `base`.`Email` ASC;
+SQL;
+    $sth = \DB::run($sql);
+    $data = $sth->fetchAll();
+    if (empty($data)) {
+        return null;
+    }
+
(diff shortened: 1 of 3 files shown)


mshappe:
Based on our e-mail thread, have you already updated the code for case insensitivity?

AricStewart:
> Based on our e-mail thread, have you already updated the code for case insensitivity?

Yup

mshappe:
You should push that code here, then. The code I'm seeing here still looks case sensitive. Or am I missing a thing?

tools/clean_duplicate_accounts.php
```diff
@@ -0,0 +1,134 @@
...
+SQL;
+    $sth = \DB::run($sql);
+    $data = $sth->fetchAll();
+    if (empty($data)) {
+        return null;
+    }
+
+    $result = [];
+    foreach ($data as $entry) {
+        $email = strtolower($entry['Email']);
```
AricStewart, line 39: Here is where I am strtolower to make the PHP parts case insensitive.

tools/clean_duplicate_accounts.php
```diff
@@ -0,0 +1,134 @@
...
+        FROM
+            `Members` AS base
+        WHERE
+            (
+            SELECT
+                COUNT(AccountID) AS COUNT
+            FROM
+                `Members`
+            WHERE
+                Email = base.Email
```
AricStewart, line 26: MySQL is by default case insensitive for string based entries.

mshappe: ...seriously?

MySQL is WEIRD. 😁  But it works in our favour here, so whatevs.

mshappe (approved):

PR #472, opened 2021-01-25
Aric/log

AricStewart:
Moves Log handling to the API and loads on demand instead of every time the admin page is loaded.


mshappe:
I need to figure out how I want to get Vue incorporated early. This may mean I need to do a focused PR on the webpack/modularization work separate from the store work. All that direct DOM manipulation in Javascript makes my eyes bleed 😁

mshappe (approved):

PR #479, opened 2021-02-08
Just the API for Stores and Products.

mshappe:
Key features of this PR:

* Introduction of [Atlas Query](http://atlasphp.io/cassini/query/), a relatively light skin over PDO and SQL, to make queries a little less onerous.
* Attempt to be fairly robust with error checking.
* Added to `BaseController` a `filterBodyParameters` method, allowing disallowed parameters to be filtered out, which in turn allows `$body` to be fed directly to certain Atlas statements for insert/update!
* Oh, and actually populate an API for stores and products :-D

This was also unit tested courtesy of Postman's testing capabilities, which are nifty. I'm still learning the ins and outs, but I believe this can eventually be fully automated and added to a CI/CD pipeline. For the moment, however, it can only be accessed within Postman. I have set up a shared team, and invited Aric to it.

### DIFF
--- a/api/src/App/Routes.php
+++ b/api/src/App/Routes.php
@@ -94,6 +94,22 @@ $app->group(
 )->add(new App\Middleware\CiabMiddleware($app))->add($authMiddleware);
 
 $app->group(
+    '/stores',
+    function () use ($app, $authMiddleware) {
+        $app->get('[/]', 'App\Controller\Stores\ListStores');
+        $app->get('/{id}', 'App\Controller\Stores\GetStore');
+        $app->post('[/]', 'App\Controller\Stores\PostStore');
+        $app->put('/{id}', 'App\Controller\Stores\PutStore');
+        $app->delete('/{id}', 'App\Controller\Stores\DeleteStore');
+        $app->get('/{store_id}/products', 'App\Controller\Stores\ListProducts');
+        $app->post('/{store_id}/products', 'App\Controller\Stores\PostProduct');
+        $app->get('/{store_id}/products/{id}', 'App\Controller\Stores\GetProduct');
+        $app->put('/{store_id}/products/{id}', 'App\Controller\Stores\PutProduct');
+        $app->delete('/{store_id}/products/{id}', 'App\Controller\Stores\DeleteProduct');
+    }
+)->add(new App\Middleware\CiabMiddleware($app))->add($authMiddleware);
+
+$app->group(
     '/event',
     function () use ($app, $authMiddleware) {
         $app->get('[/]', 'App\Controller\Event\ListEvents');
--- a/api/src/Controller/BaseController.php
+++ b/api/src/Controller/BaseController.php
@@ -5,6 +5,8 @@
 
 namespace App\Controller;
 
+use ArrayObject;
+
 use Slim\Container;
 use Slim\Http\Response;
 use Slim\Http\Request;
@@ -92,12 +94,17 @@ abstract class BaseController
         } elseif ($type == BaseController::RESULT_TYPE) {
             return $data;
         } else {
+            if (count($result) == 3) {
+                $code = $result[2];
+            } else {
+                $code = 200;
+            }
             $includes = $request->getQueryParam('include', null);
             if ($includes) {
                 $values = array_map('trim', explode(',', $includes));
                 $this->processIncludes($request, $response, $args, $values, $data);
             }
-            return $this->jsonResponse($request, $response, $data);
+            return $this->jsonResponse($request, $response, $data, $code);
         }
 
     }
@@ -145,6 +152,19 @@ abstract class BaseController
             }
         }
         return $data;
+
+    }
+
+
+    protected function filterBodyParams(array $permitted_keys, array $body): array
+    {
+        $ret = (new ArrayObject($body))->getArrayCopy();
+        $diff = array_diff(array_keys($body), $permitted_keys);
+        foreach ($diff as $key) {
+            unset($ret[$key]);
+        }
+
+        return $ret;
 
     }
 
@@ -343,6 +363,23 @@ abstract class BaseController
         }
         $data = BaseController::mapMemberData($data['users'][0]);
         return $data;
+
+    }
+
+    
+    public function notFoundResponse(
+        Request $request,
+        Response $response,
+        String $type,
+        string $key
+    ): Response {
+        return $this->errorResponse(
+            $request,
+            $response,
+            "Could not find $type ID $key",
+            'Not Found',
+            404
+        );
 
     }
 
--- a/api/src/Controller/Stores/BaseProduct.php
+++ b/api/src/Controller/Stores/BaseProduct.php
@@ -1,0 +1,60 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller\Stores;
+
+use Atlas\Query\Select;
+
+use Slim\Container;
+use Slim\Http\Request;
+use Slim\Http\Response;
+use App\Controller\BaseController;
+
+abstract class BaseProduct extends BaseController
+{
+    
+    /**
+     * @var int
+     */
+    protected $id = 0;
+
+
+    public function __construct(Container $container)
+    {
+        parent::__construct('products', $container);
+        
+    }
+
+
+    protected function buildProductHateoas(Request $request, Number $store_id)
+    {
+        if ($this->id !== 0) {
+            $path = $request->getUri()->getBaseUrl();
+            $this->addHateoasLink('self', $path.'/stores/'.strval($store_id).'/products/'.strval($this->id), 'GET');
+        }
+
+    }
+
+
+    protected function getProduct(array $params, Request $request, Response $response, &$error)
+    {
+        $select = Select::new($this->container->db);
+        $select->columns('ProductID as id', 'StoreID', 'Name', 'ProductSlug', 'Description', 'UnitPriceCents', 'PaymentSystemRef');
+        $select->from('Products')->whereEquals(['ProductID' => $params['id']]);
+        $product = $select->fetchOne();
+
+        if (empty($product)) {
+            $error = [
+            BaseController::RESULT_TYPE,
+            $this->errorResponse($request, $response, "Could not find Product with ID ${params['id']}", 'Not found', 404)
+            ];
+            
+            return null;
+        }
+
+        return $product;
+
+    }
+
+
+    /* End BaseProduct */
+}
--- a/api/src/Controller/Stores/DeleteProduct.php
+++ b/api/src/Controller/Stores/DeleteProduct.php
@@ -1,0 +1,44 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller\Stores;
+
+use App\Controller\BaseController;
+
+use Atlas\Query\Delete;
+
+use Slim\Http\Request;
+use Slim\Http\Response;
+
+class DeleteProduct extends BaseProduct
+{
+
+
+    public function buildResource(Request $request, Response $response, $params): array
+    {
+        if (!$_SESSION['IS_ADMIN']) {
+            return [
+            BaseController::RESULT_TYPE,
+            $this->errorResponse($request, $response, 'Permission Denied', 'Permission Denied', 403)];
+        }
+
+        $delete = Delete::new($this->container->db);
+        $result = $delete->from('Products')->whereEquals(['ProductID' => $params['id']])->perform();
+
+        if ($result->rowCount() == 0) {
+            return [
+            BaseController::RESULT_TYPE,
+            $this->errorResponse($request, $response, 'Already deleted', 'Not found', 404)
+            ];
+        }
+
+        return [
+        BaseController::RESOURCE_TYPE,
+        [null],
+        204
+        ];
+
+    }
+
+
+    /* end DeleteProduct */
+}
--- a/api/src/Controller/Stores/DeleteStore.php
+++ b/api/src/Controller/Stores/DeleteStore.php
@@ -1,0 +1,44 @@
+<?php declare(strict_types=1);
+
+namespace App\Controller\Stores;
+
+use App\Controller\BaseController;
+
+use Atlas\Query\Delete;
+
+use Slim\Http\Request;
+use Slim\Http\Response;
+
+class DeleteStore extends BaseStore
+{
+
+
+    public function buildResource(Request $request, Response $response, $params): array
+    {
+        if (!$_SESSION['IS_ADMIN']) {
+            return [
+            BaseController::RESULT_TYPE,
+            $this->errorResponse($request, $response, 'Permission Denied', 'Permission Denied', 403)];
+        }
+        
+        $delete = Delete::new($this->container->db);
+        $result = $delete->from('Stores')->whereEquals(['StoreID' => $params['id']])->perform();
+
+        if ($result->rowCount() == 0) {
+            return [
+            BaseController::RESULT_TYPE,
+            $this->errorResponse($request, $response, 'Already deleted', 'Not Found', 404)
+            ];
+        }
+
+        return [
+        BaseController::RESOURCE_TYPE,
+        [null],
+        204
+        ];
+
+    }
+
+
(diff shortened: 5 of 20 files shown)


api/index.php
```diff
@@ -1,5 +1,10 @@
 <?php
 
 require __DIR__.'/src/App/App.php';
-
+error_reporting(E_ALL);
+set_error_handler(function ($severity, $message, $file, $line) {
+    if (error_reporting() & $severity) {
+        throw new \ErrorException($message, 0, $severity, $file, $line);
+    }
+});
```
AricStewart, line 9: Ok I am going to be picky...

I would love all changed in a given commit to be related.  This seems valid and all but is not directly related to the store code later. 

So my gut says a different commit maybe different pull request.

api/src/App/Routes.php
```diff
@@ -93,6 +93,38 @@ function () use ($app, $authMiddleware) {
...
+    '/payment',
+    function () use ($app, $authMiddleware) {
+        $app->add(function ($request, $response, $next) {
+            \Stripe\Stripe::setApiKey($_ENV['STRIPE_PRIVATE_KEY']);
+            return $next($request, $response);
+        });
+        $app->post('/checkout', 'App\Controller\Payment\PostCheckout');
+    }
+)->add(new App\Middleware\CiabMiddleware($app))->add($authMiddleware);
+
```
AricStewart, line 106: We talked about moving the payment into a vendor driven sort of framework.

Is that a version next?

api/src/Controller/BaseController.php
```diff
@@ -148,6 +155,18 @@ protected function filterOutput(Request $request, $data, $code): array
 
     }
```
AricStewart, line 156: Could you explain the need for these BaseController changes? 
You want to send a code?

api/src/Controller/Member/GetStatus.php
```diff
@@ -65,6 +65,7 @@ public function buildResource(Request $request, Response $response, $args): arra
         }
         $data = $data['users'][0];
         if (!array_key_exists('id', $data)) {
+            $error = 'User ID not found';
```
AricStewart, line 68: Is this relevant to the rest of the changes or is this just something you found and wanted to fix as well?

api/src/App/Routes.php
```diff
@@ -93,6 +93,38 @@ function () use ($app, $authMiddleware) {
...
+        $app->get('[/]', 'App\Controller\Stores\ListStores');
+        $app->get('/{id}', 'App\Controller\Stores\GetStore');
+        $app->post('[/]', 'App\Controller\Stores\PostStore');
+        $app->put('/{id}', 'App\Controller\Stores\PutStore');
+        $app->delete('/{id}', 'App\Controller\Stores\DeleteStore');
+        $app->get('/{store_id}/products', 'App\Controller\Stores\ListProducts');
+        $app->post('/{store_id}/products', 'App\Controller\Stores\PostProduct');
+        $app->get('/{store_id}/products/{id}', 'App\Controller\Stores\GetProduct');
+        $app->put('/{store_id}/products/{id}', 'App\Controller\Stores\PutProduct');
+        $app->delete('/{store_id}/products/{id}', 'App\Controller\Stores\DeleteProduct');
```
AricStewart, line 124: hummmmmm. I can go either way here... But me current standard is to not use the plural of the resource. 
member, department, etc... so product**s**  and store**s** is a bit of a break from that.

api/src/Controller/Stores/BaseProduct.php
```diff
@@ -0,0 +1,58 @@
...
+    
+    /**
+     * @var int
+     */
+    protected $id = 0;
+
+
+    public function __construct(Container $container)
+    {
+        parent::__construct('products', $container);
```
AricStewart, line 23: plural again.

api/src/Controller/Stores/BaseProduct.php
```diff
@@ -0,0 +1,58 @@
...
+        $product = $select->fetchOne();
+
+        if (empty($product)) {
+            return [
+            BaseController::RESULT_TYPE,
+            $this->errorResponse($request, $response, "Could not find Product with ID ${params['id']}", 'Not found', 404)
+            ];
+        }
+
+        return null;
```
AricStewart, line 52: oh, humm.  I will have to think this design choice.
You are using an argument as a return and returning null.
The rest of the code all returns objects vs arrays and uses is_array and such to determine if it is an error or not.

api/src/Controller/Stores/DeleteProduct.php
```diff
@@ -0,0 +1,43 @@
...
+use Slim\Http\Request;
+use Slim\Http\Response;
+
+class DeleteProduct extends BaseProduct
+{
+
+
+    public function buildResource(Request $request, Response $response, $params): array
+    {
+        if (!$_SESSION['IS_ADMIN']) {
```
AricStewart, line 18: Generally I have been liking using RBAC methods here..
`        if (\ciab\RBAC::havePermission('api.delete.store'))`
Relying on `IS_ADMIN` is not the direction I think we should keep going.

api/src/Controller/Stores/DeleteProduct.php
```diff
@@ -0,0 +1,43 @@
...
+            return [
+            BaseController::RESULT_TYPE,
+            $this->errorResponse($request, $response, 'Permission Denied', 'Permission Denied', 403)];
+        }
+
+        $error = $this->getProduct($request, $response, $product, $params);
```
AricStewart, line 24: Feels a git complicated to just verify there is a row to delete.

The way (not Atlas) I have been doing it now is
```
        $sql = "DELETE FROM `Registrations` WHERE RegistrationID = $id";
        $sth = $this->container->db->prepare($sql);
        $sth->execute();
        if ($sth->rowCount() == 0) {
```

api/src/Controller/Stores/DeleteStore.php
```diff
@@ -0,0 +1,43 @@
...
+        
+        $error = $this->getStore($request, $response, $store, $params);
+
+        if (!empty($error)) {
+            return $error;
+        }
+
+        $delete = Delete::new($this->container->db);
+        $delete->from('Stores')->whereEquals(['StoreID' => $params['id']])->perform();
+
```
AricStewart, line 32: See above

api/src/Controller/Stores/GetProduct.php
```diff
@@ -0,0 +1,38 @@
...
+use App\Controller\BaseController;
+
+class GetProduct extends BaseProduct
+{
+
+
+    public function buildResource(Request $request, Response $response, $params): array
+    {
+        // TODO: RBAC more than just admin maybe
+        if (!$_SESSION['IS_ADMIN']) {
```
AricStewart, line 17: IS_ADMIN

api/src/Controller/Stores/PostProduct.php
```diff
@@ -0,0 +1,87 @@
...
+use Slim\Http\Response;
+use App\Controller\BaseController;
+
+class PostProduct extends BaseProduct
+{
+
+
+    public function buildResource(Request $request, Response $response, $params): array
+    {
+        // TODO: RBAC more than just admin maybe
```
AricStewart, line 17: Yup!

api/src/Controller/Stores/PostStore.php
```diff
@@ -0,0 +1,61 @@
...
+use Slim\Http\Request;
+use Slim\Http\Response;
+
+class PostStore extends BaseStore
+{
+
+
+    public function buildResource(Request $request, Response $response, $args): array
+    {
+        // TODO: RBAC more than just admin maybe
```
AricStewart, line 15: ditto

docker/php/Dockerfile
```diff
@@ -1,8 +1,11 @@
 FROM php:7.2.7-fpm-alpine3.7
 RUN apk update \
     && apk upgrade \
-    && apk add git bash msmtp freetype libpng libjpeg-turbo freetype-dev libpng-dev libjpeg-turbo-dev;
+    && apk add git bash msmtp freetype libpng libjpeg-turbo freetype-dev libpng-dev libjpeg-turbo-dev autoconf build-base;
 RUN docker-php-ext-install mysqli pdo_mysql
 RUN docker-php-ext-install gd
+RUN pecl install apfd
+RUN echo 'extension=apfd.so' > /usr/local/etc/php/conf.d/apfd.ini
+RUN mkdir -p /tmp/scss_cache/scss && chmod -R 0777 /tmp/scss_cache
```
AricStewart, line 9: Confusing a few different changes for different reasons.

Is the scss_cache change still needed?

modules/store/database-schema.inc
```diff
@@ -0,0 +1,66 @@
...
+    ],
+    'PurchaseLineItems' => [
+        'PurchaseID' => 'Purchases (PurchaseID) ON DELETE RESTRICT ON UPDATE CASCADE',
+        'ProductID' => 'Products (ProductID) ON DELETE RESTRICT ON UPDATE CASCADE',
+    ]
+];
+
+$MODULE_TABLES = array_merge($MODULE_TABLES, $DB_tables);
+$MODULE_FOREIGNKEYS = array_merge($MODULE_FOREIGNKEYS, $DB_foreignKeys);
+$MODULE_INDEX = array_merge($MODULE_INDEX, $DB_index);
```
AricStewart, line 66: SonarCloud really hates that we have a lot of things named `$DB_tables `etc... so I have  a pull request to specifically uniquely rename all those bits.
so `$store_DB_tables`   etc...

api/src/App/Routes.php
mshappe: Yeah, the rest of the payment code isn't even in here, honestly -- I held it out. So this section can just get cut for now.

api/src/Controller/BaseController.php
mshappe: The correct code for a successful Create is 201 (unless it's pushed to a queue, in which case it's 202).

The correct code for a successful Delete is 204.

There was no way to send that without this change.

api/src/App/Routes.php
mshappe: The Internet-wide standard, however, is plural.

api/src/Controller/Member/GetStatus.php
mshappe: Since $error is otherwise undefined, the rest of this clause is invalid without this.

api/src/Controller/Stores/BaseProduct.php
mshappe: And staying that way :-) The others should all be corrected.

api/src/Controller/Stores/BaseProduct.php
mshappe: This is actually based direct on `getCycle`, although that's otherwise broken at the moment (it doesn't get request and response passed in, so it will 500 when it reaches the equivalent point.

api/src/Controller/Stores/DeleteProduct.php
mshappe: I agree, but I also don't want RBAC to get too crazy. It will be unmanageable if it continues to be based on fine-grained permissions like that. RBAC is about ROLE.

api/src/Controller/Stores/DeleteProduct.php
mshappe: The row-count trick would also work here, I believe, so that's possible.

docker/php/Dockerfile
mshappe: I haven't had a chance to test for that.

mshappe: Also, the APFD change is something of a mixed bag that I still need to exercise a bit more as to whether I really need it for what I'm doing here. PHP has a bad defect that they refuse to correct, insisting that PUT parameters should always been in the URL (they should not be). However, the main place it really breaks is when paired with `FormData`, which I'd started to use in Javascript. Now, however, I'm mainly using JSON instead all the way. I need to test whether apfd is needed for that. If it isn't, this whole change can go away.

api/src/App/Routes.php
AricStewart: Sounds good. Lets cut this code and put it in the next chunk.

api/src/Controller/BaseController.php
AricStewart: Wow so I totally did not know about and of those 20x codes.  Thanks! The rest of the code will have to be updated...

api/src/Controller/Member/GetStatus.php
AricStewart: so a bug fix!  Thank!  Feels like a new commit maybe a new PR, but honestly for a small commit like this just make the one line fix, give it a good git commit description and push it.

api/src/App/Routes.php
AricStewart: really, How did i miss that in my restful api research.  Again my restful api research was mostly about 3-4 years ago. Is there a standards docuemt somewhere I should read? Well sounds like we need to update all the rest of the API.

docker/php/Dockerfile
AricStewart: Honestly JSON should be our standard.  It is not hard to convert FormData into JSON.

mshappe: Agreed. The spike PR is not currently up to date, but my spike branch already made this change (including a small update to `apiRequest` to fully support it).

api/src/App/Routes.php
mshappe: I'm not sure I've ever seen it written down, and it's always possible it's more a Rails-specific interpretation of RESTFUL conventions than any formal standard, but in 10 years doing this, I've always seen these as plural.

api/src/Controller/BaseController.php
mshappe: Yeah, it's not a huge emergency when we're talking to ourselves and checking only against our own expectations, but I know we're trying to be Good Citizens here. I'm not sure how picky other clients would be; only that 201 means 'Created' and should be what's sent on a Create post. Any OTHER use of POST (say, a form that results in sending e-mail and not creating a record) could return 200, so it's not so much specific to POST as specific to something being created as a result of the action.

api/src/Controller/Stores/BaseProduct.php
mshappe: Point being, I assumed there was a Reason we were doing it that way (in `getCycle()`) and followed suit. 

I don't love testing for array vs object as a way of determining error, but I do feel the semantics here should be inverted: return the sought object, or else null, and if null, set `&$error` with the array. Something called `getThing` should return Thing or nothing.

docker/php/Dockerfile
mshappe: These changes appear not to be need for JSON payloads, so I'm withdrawing them unless something else comes along that seems to need them.

api/src/Controller/BaseController.php
```diff
@@ -346,6 +366,23 @@ public function findMemberId(
...
 
+    
+    public function notFoundResponse(
```
mshappe, line 370: I'm leaving this for now, even though I'm not using it as much as I thought I would. Really, what I want to do is add an exception handling in `__invoke` so we can just throw `NotFound` exceptions and get rid of a lot of repeated code. But that's DEFINITELY another PR.

api/src/App/Routes.php
AricStewart: ok we can fight that fight however. But I do crave consistency. The open source developer in my says "Always conform to the standards of the existing product" But I will cut you some slack here.

api/src/Controller/BaseController.php
AricStewart: I like that idea. Let move toward that.

api/src/Controller/Stores/DeleteProduct.php
AricStewart: This is in need of a broader conversation.  Here directly because I really dislike the _SESSION['IS_ADMIN'] and have been actively working to remove it everything.  so ADDING instances makes me unhappy.

AricStewart (approved):

PR #482, opened 2021-02-12
Introduce the idea of a NotFoundException handler

mshappe:
Allow 404 handling to be centralized and DRY'd up by having get methods throw a `NotFoundException` instead of having to handle it locally.

### DIFF
--- a/api/src/Controller/BaseController.php
+++ b/api/src/Controller/BaseController.php
@@ -6,6 +6,7 @@
 namespace App\Controller;
 
 use ArrayObject;
+use Exception;
 
 use Slim\Container;
 use Slim\Http\Response;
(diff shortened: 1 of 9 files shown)


api/src/Controller/Stores/PutProduct.php
```diff
@@ -23,10 +23,6 @@ public function buildResource(Request $request, Response $response, $params): ar
 
         $product = $this->getProduct($params, $request, $response, $error);
```
mshappe, line 24: SonarCloud rightly bitches about this being a useless assignment. It actually occurs to me that we can use the `rowCount` trick here and `PutStore`, too (if the update affects 0 rows it's a 404). So I'll be updating with that, maybe later tonight.

api/src/Controller/BaseController.php
```diff
@@ -75,7 +81,15 @@ public function processIncludes(Request $request, Response $response, $args, $in
 
     public function __invoke(Request $request, Response $response, $args)
     {
-        $result = $this->buildResource($request, $response, $args);
+        try {
```
mshappe, line 84: SonarCloud also complains about the complexity of this method now. I see a way to refactor it, and will tackle that later.

AricStewart (approved):

PR #488, opened 2021-02-14
Implement InvalidParameterException and ConflictException

AricStewart:


api/src/Controller/Announcement/PostAnnouncement.php
```diff
@@ -124,14 +125,10 @@ public function buildResource(Request $request, Response $response, $args): arra
 
         $body = $request->getParsedBody();
         if (!array_key_exists('Scope', $body)) {
-            return [
-            \App\Controller\BaseController::RESULT_TYPE,
-            $this->errorResponse($request, $response, 'Required \'Scope\' parameter not present', 'Missing Parameter', 400)];
+            throw new InvalidParameterException('Required \'Scope\' parameter not present');
```
mshappe, line 128: I'm going to include some "DRY thoughts" in my commentary to this PR, which do not need to be acted on right now, but would be good to think about for another pass.

DRY thought:
```php
$mandatory_keys = ['Scope', 'Text'];
foreach ($mandatory_keys as $key) {
  if (!array_key_exists($key, $body)) {
    throw  new InvalidParameterException(...)
  }
}
```

And then, as a further refinement, factor that up to the base class as `validateMandatoryParams`, so the invocation here becomes
```php
$this->validateMandatoryParams(['Scope', 'Text']);
```

api/src/Controller/BaseController.php
```diff
@@ -99,6 +111,16 @@ public function __invoke(Request $request, Response $response, $args)
             BaseController::RESULT_TYPE,
             $this->errorResponse($request, $response, $e->getMessage(), 'Permission Denied', 403)
             ];
+        } catch (InvalidParameterException $e) {
```
mshappe, line 114: DRY thought:

```php
$CODE_MESSAGES = [ 403 => 'Permission denied', 404 => `'Not Found' ]; /* etc */

$code = 200;
try {
  something();
} catch (foo e) {
  $code = 404;
} catch (bar e) {
  $code = 403
} finally {
  if ($code >= 400) {
    $result = [
    BaseController::RESULT_TYPE,
    $this->errorResponse(blah, blah, $CODE_MESSAGES[$code], $code);
    ];
  }
}
```

mshappe (approved):

PR #491, opened 2021-02-15
Aric/testing

AricStewart:


api/tests/announcement_test.py
```diff
@@ -0,0 +1,228 @@
...
+    """ Getting the announcments as a privilaged user """
+
+    target = add_announcement()
+
+    response = requests.get(SERVER+"/api/department/1/announcements",
+                   headers={'Authorization': 'Bearer '+ADMIN_ACCESS},
+                   data={})
+    assert_status_code(response, 200)
+    data = response.json()
+    assert 'data' in data
```
mshappe, line 66: The more usual way to test that a response body has been fulfilled is to map them both to hashes/dictionaries and compare those, using whatever comparison operations the language provides.

So, you'd say something like:
```python
expected_foo = {
  bar: 'baz',
  quux: 'quuux',
  zorch: {
    spaz: 'squid'
  }
}

response = requests.get(...)
data = response.json()

# whatever you do to compare hashes in Python goes here
```

mshappe (commented):
Before we go much further, we should decide if we really want to do this in Python or not. I know we agreed it was OK, but the Codeception test framework I linked yesterday looks pretty solid, and is worth considering.

Also, see my comment about drying up the response-body assertions.

api/tests/announcement_test.py
AricStewart: I agree the long list of asserts sucks and a better way to frame it would be welcome.

Some web searching says that this can be messy, because json can be reordered, and if I am reading this correctly we could get into trouble with that (https://stackoverflow.com/questions/25851183/how-to-compare-two-json-objects-with-the-same-elements-in-a-different-order-equa)

Secondly there are fields we do not care about and do not want to compare, date fields spring to mind. So we would have to handle those fields.   It just makes the == a much harder operation than would be hoped. 

Unless you know magic to fix that. :)

AricStewart:
> Before we go much further, we should decide if we really want to do this in Python or not. I know we agreed it was OK, but the Codeception test framework I linked yesterday looks pretty solid, and is worth considering.

Ok we should talk since I thought pytest was the way to go and I spent the morning learning that. :) I am personally not very much a fan of frameworks unless you can argue just what benefit we get out of having to add their dependency to the project. 

Codeception is composer'able  so that is a benefit.  But it would mean writing our test in PHP. What benefits do you see from Codeception than pytest?

api/tests/announcement_test.py
mshappe: AFAIK, python set operators don't care about order, and dictionary comparisons of keys and values are treated as set comparisons.

```python
>>> {'bar': 'foo', 'baz': 'quux'} == {'baz': 'quux', 'bar': 'foo'}                                                      
True 
```

The same is true if you just compare keys:

```python
>>> {'bar': 'foo', 'baz': 'quux'}.keys() == {'baz': 'quux', 'bar': 'foo'}.keys()
True 
```

So the real question is, what are we testing here. If we're testing structure, we can use `keys()` to do so and still get down to not having to line-by-line compare. If it's values we care about, we should be unit testing the functions that return the values. If we simply can't do that, then we should be testing values we've inserted into the database ourselves, as the 'before' part of a test, which means we can in fact compare the dates with accuracy, because they'll be dates we just inserted explicitly.

PR #494, opened 2021-02-20
Introduce Webpack, begin modularizing JavaScript

mshappe:
* Introduces Webpack, Babel, and Vue (not used yet)
* Converts the newer JS supporting Registration to ES6+ modules
* Serves those modules via Webpack

I chose registration as an initial target because it was already half-way there (livin' on a prayer!) I also realized in general I needed to start from the BOTTOM level and work my way UPWARD to common.js if I wanted to do this gradually, rather than change everything everywhere all at once!

Webpack is set up to transpile via Babel

Babel is set up to transpile to its default presets, meaning it will recognize anything up to the very minute and transpile back to ES5. If we need to go further back than that, that's an option, it's just not configured right now.

### DIFF
--- a/modules/registration/sitesupport/registration.js
+++ b/modules/registration/sitesupport/registration.js
@@ -3,6 +3,7 @@
  */
 
 /* jshint browser: true */
+/* jshint esversion: 9 */
 /* jshint -W097 */
 /* global confirmbox, showSidebar, basicBackendRequest, showSpinner,
           hideSpinner, apiRequest */
@@ -15,7 +16,7 @@
 var sidebarMainDiv = 'main_content';
 
 function basicRegistrationRequest(parameter, finish) {
-  basicBackendRequest('POST', 'registration', parameter, finish);
+  basicBackendRequest('GET', 'registration', parameter, finish);
 }
 
 function refreshBadgeData(badge) {
--- a/package.json
+++ b/package.json
@@ -1,12 +1,35 @@
 {
-  "name": "CIAB-Portal",
+  "name": "ciab-portal",
   "version": "1.3.0",
   "devDependencies": {
+    "@babel/core": "^7.12.17",
+    "@babel/preset-env": "^7.12.17",
+    "babel-loader": "^8.2.2",
+    "clean-webpack-plugin": "^3.0.0",
     "eslint": "^6.6.0",
     "eslint-plugin-lodash": "^1.10.1",
-    "eslint-plugin-node": "^2.0.0"
+    "eslint-plugin-node": "^2.0.0",
+    "html-webpack-plugin": "^4.5.2",
+    "webpack": "^5.23.0",
+    "webpack-cli": "^4.5.0",
+    "webpack-dev-server": "^3.11.2",
+    "webpack-merge": "^5.7.3"
   },
   "scripts": {
-    "lint": "node_modules/eslint/bin/eslint.js ."
+    "lint": "node_modules/eslint/bin/eslint.js .",
+    "start": "nodemon --watch webpack.dev.js --watch webpack.common.js --exec 'webpack serve --config webpack.dev.js'",
+    "build": "webpack --config webpack.prod.js"
+  },
+  "babel": {
+    "presets": [
+      "@babel/preset-env"
+    ]
+  },
+  "dependencies": {
+    "glob": "^7.1.6",
+    "nodemon": "^2.0.7",
+    "vue": "^2.6.12",
+    "vue-loader": "^15.9.6",
+    "vue-template-compiler": "^2.6.12"
   }
 }
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "@babel/preset-env": "^7.12.17",
     "babel-loader": "^8.2.2",
     "clean-webpack-plugin": "^3.0.0",
-    "eslint": "^6.6.0",
+    "eslint": "^6.8.0",
     "eslint-plugin-lodash": "^1.10.1",
     "eslint-plugin-node": "^2.0.0",
     "html-webpack-plugin": "^4.5.2",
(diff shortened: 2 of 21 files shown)


mshappe:
Note: ask lots of questions if you need to. This is the beginning of a sweeping change, and I'm not 100% certain of my approach yet, so I'm wide open to discussion here.

package.json
```diff
@@ -1,12 +1,35 @@
 {
-  "name": "CIAB-Portal",
+  "name": "ciab-portal",
```
AricStewart, line 2: Why? Mixed Case bugging you too much?

modules/registration/sitesupport/registration.js
```diff
@@ -15,7 +16,7 @@
 var sidebarMainDiv = 'main_content';
 
 function basicRegistrationRequest(parameter, finish) {
-  basicBackendRequest('POST', 'registration', parameter, finish);
+  basicBackendRequest('GET', 'registration', parameter, finish);
```
AricStewart, line 19: THis one will be going away soon anyway but WHY?  The request using this is  a POST request.

pages/head/public.inc
```diff
@@ -1,5 +1,4 @@
 <meta http-equiv="Cache-control" content="No-Cache">
-<script src="sitesupport/common.js"></script>
```
AricStewart, line 2: Not needed?

sitesupport/configuration.js
```diff
@@ -115,7 +115,7 @@ var configHeader = function(options) {
 }
 
 
-var settingsTable = function(options) {
+export var settingsTable = function(options) {
```
AricStewart, line 118: We are not a module yet here... Does this work outside of module code?

webpack.common.js
```diff
@@ -0,0 +1,81 @@
+/* jshint esversion: 9 */
+
+const path = require('path');
```
AricStewart, line 3: This file is hard to read with the eslint errors.  
Could you add a /* globals ...  */ statement to the top to silence those note defined  messages please?

Then have eslint installed (it is a node.js tool info is on the main README) and you can even do a eslint --fix to fix all the whitespace and indenting problems automagically.

package.json
mshappe: Linter appeasement. The "standard" for package names is all lowercase.

(VSCode lints and highlights both PHP and Javascript for me now, on the fly, with the PHP mostly obeying the same rules we've got the GitHub linter using).

modules/registration/sitesupport/registration.js
mshappe: You'd think so, but I got a 500 trying it that way!

mshappe: (This can be reverted if we can figure out what was really going on)

pages/head/public.inc
mshappe: We already include common.js in the top-level head-before that every single page gets! We've got a lot of redundancies like these that I'm picking off along the way.

sitesupport/configuration.js
mshappe: I haven't seen an issue yet, treating this like a module, but I can do some more digging to make certain. I think the only things that actually use it are already in code that are either deliberately modularized or close enough for jazz, but I'll double-check it.

webpack.common.js
mshappe: I'm pretty sure VScode is using eslint, and reports only one questionmark for this file -- the use of require rather than import. But the formatting of this file, even for Webpack 5, is still an "old-style" module that predates import/export.

mshappe: (Make sure that your eslint config, in turn, is not still making really old assumptions. There won't be ANY globals going forward. Modules have no globals!)

mshappe: I just deleted my comment about .eslintrc.js because I see it's actually pretty modern in its assumptions, now I wonder why you're seeing errors I don't. jshint thinks the file is clean for example.

mshappe: I think the problem is that .eslintrc.js isn't currently configured for Node.js files (which webpack's configs are). I'll look into the right settings.

mshappe: OK, I was clearly not configured to use eslint as I thought i was, so I'm bringing that inline and taking some of the recommendations into account.

One I want to change though: 80 characters is ridiculous in this day and age. I prefer 100. Thoughts?

mshappe: Also, the trailing-space thing was something that mystified me, because I thought I had configured VScode to strip them. Turns out that's only true on my work machine. Derp.

sitesupport/configuration.js
mshappe: Yeah, digging around, I may have to revert this one for the moment.

mshappe:
We actually have a lot of javascript that contains no real exports, just code that autoloads and runs immediately. I think that's fine, but I want to dig around and see how that kind of thing is handled in an ES6-9 universe--if that's still the Done Thing, or if there's a new way.

modules/registration/sitesupport/registration.js
AricStewart: Well this is only used by `refreshBadgeData`  and that is setting `refreshData` and if you look in `modules/registration/pages/pre/inc` you can see that `refreshData` is used in `POST` requests.

Now it may not be working... but that would be unrelated to it being POST vs GET. I feel like this change should be removed and in a different patch all that code can be removed.  In fact I ahve an in progress tree that is doing just that.

sitesupport/configuration.js
AricStewart: Ok tell me what you come up with one way or the other.  Feels like reverting the export seems like the right things.

AricStewart:
Inactive and will likely require rework to integrate

PR #497, opened 2021-02-23
Aric/phptest

AricStewart:


api/src/Tests/Traits/AppTestTrait.php
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+
+    /* End */
+}
+
+/**
+ * App Test Trait.
+ */
+trait AppTestTrait
```
mshappe, line 44: TIL!

(Rails would call this a Module, same basic behaviour. Beautiful).

mshappe:
This is excellent! I particularly like your leveraging of PHP `trait`s, which I didn't know about until today, to DRY up the baseline setup of tests!

I am looking forward to wiring up a CI/CD pipeline to run these!

api/src/Tests/Traits/AppTestTrait.php
AricStewart: yay!

AricStewart:
Agreed. I am trying to hook the SonarCloud report to it but it seems to be unable to find the coverage report. I will keep poking.

mshappe (dismissed):

PR #498, opened 2021-02-24
Deglobalize DB schema init

mshappe:
Also remove the lingering use of $GLOBALS['DBNAME'] in favour of $_ENV['DBNAME'] as everywhere else.


mshappe:
This is just a first pass. I actually want to rig it so that we're passing in an instance of `DB`  to setup functions, like we've talked about setting up routes for testing, but I wanted a quick proof of concept to eliminate the warnings I was seeing when running coverage.

AricStewart (approved):

PR #499, opened 2021-02-24
Remove a few more blind GLOBAL use.

AricStewart:
Cleans up the PHPTest output

### DIFF
--- a/functions/database.inc
+++ b/functions/database.inc
@@ -250,24 +250,25 @@ SQL;
 
     public function loadModules()
     {
-        global $MODULESDIR;
-
         if ($this->loaded) {
             return;
         }
-        try {
-            $modules = scandir($MODULESDIR);
-            foreach ($modules as $key => $value) {
-                if (!in_array($value, array(".", ".."))) {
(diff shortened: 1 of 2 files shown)


functions/database.inc
```diff
@@ -250,24 +250,25 @@ SQL;
...
             return;
         }
-        try {
-            $modules = scandir($MODULESDIR);
-            foreach ($modules as $key => $value) {
-                if (!in_array($value, array(".", ".."))) {
-                    if (is_dir($MODULESDIR.DIRECTORY_SEPARATOR.$value)) {
-                        if (is_file($MODULESDIR.DIRECTORY_SEPARATOR.$value.DIRECTORY_SEPARATOR.'database-schema.inc')) {
-                            require_once($MODULESDIR.DIRECTORY_SEPARATOR.$value.DIRECTORY_SEPARATOR.'database-schema.inc');
+        if (array_key_exists('MODULESDIR', $GLOBALS) &&
```
mshappe, line 256: I agree with this change, but eventually, I'd like to de-globalize everything. Also, if we ever want to unit-test the modules, we'll need to load them! But that can be solved by having the test suite set MODULESDIR, so it's not the world.

mshappe (approved):

PR #500, opened 2021-02-26
Foreign keys are working now, so this has to be done in a particular …

mshappe:
…order

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,12 @@ vendor
 #################
 .env
 .vscode
+
+# PHPUnit #
+###########
+build/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -52,3 +52,12 @@ vendor
...
 .vscode
+
+# PHPUnit #
+###########
+build/
+.phpunit.result.cache
+
+# Javascript/Node stuff #
+#########################
+node_modules/
```
AricStewart, line 63: You know, this is an unrelated change... I am going to let it slide but I have to comment...

AricStewart (approved):

PR #507, opened 2021-02-27
Aric/concom resource

AricStewart:


api/src/Modules/concom/Controller/BaseConcom.php
```diff
@@ -21,11 +21,75 @@ public function __construct(Container $container)
...
+    protected function getConComPosition($account, $event = null)
+    {
+        if ($event == null) {
+            $event = \current_eventID();
+            if ($event === null) {
+                return array();
+            }
+        }
+        $sql = <<<SQL
+            SELECT
```
mshappe, line 33: I was about to ask if I could convince you to start using Atlas\Query classes, but while I'm sure I could write this query with Atlas, I don't know exactly how I'd do it, yet! 😄 

That said, I'd appreciate it if you'd start using it for more common queries (http://atlasphp.io/cassini/query/ , and lots of examples now in Stores API and the new test_prereqs). I want to start getting out of the habit of writing raw SQL as much as possible.

api/src/Tests/Cases/DeadlineTest.php
```diff
@@ -59,12 +65,25 @@ protected function setUp(): void
 
         $this->target = $target;
 
+        $when = date('Y-m-d', strtotime('+1998 years'));
+        $to = date('Y-m-d', strtotime('+2002 years'));
+        $this->cycle = $this->runSuccessJsonRequest('POST', '/cycle', null, ['From' => $when, 'To' => $to], 201);
+        $when = date('Y-m-d', strtotime('+1999 years'));
+        $to = date('Y-m-d', strtotime('+1999 years'));
+        $this->event = $this->runSuccessJsonRequest('POST', '/event', null, ['From' => $when, 'To' => $to, 'Name' => 'PHPTest-a-con'], 201);
```
mshappe, line 73: So, this is an example of what I mean by "just use the database". We're not testing the Event API here, but the Deadline API. So instead of using the Event API, just us an Atlas\Query\Insert to stash the objects and an Atlas\Query\Delete to tear them down later.

mshappe (approved):
I've made comments here, but they're not show-stoppers, so I'm approving it.

PR #512, opened 2021-03-01
Use OAuth2 in testing

AricStewart:


api/src/Tests/Cases/OAuth2Test.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        $this->setupToken = false;
+        parent::setUp();
+
+    }
+
+
+    public function testTokens(): void
+    {
+        $this->runRequest('POST', '/token', null, ['grant_type' => 'passwordx', 'username' => self::$login, 'password' => self::$password, 'client_id' => self::$client], 400);
```
mshappe, line 21: One trick: look at PHPUnit's docco into data providers to DRY this up a little. It's nifty.

mshappe (dismissed):
See my note about using data provider methods, but other than that, this looks fine. I'm not sure what about the password part itches you, but this looks normal to me.

api/src/Tests/Cases/OAuth2Test.php
AricStewart: Provider figured out and used

api/src/Tests/Cases/OAuth2Test.php
```diff
@@ -0,0 +1,59 @@
...
+
+    protected function setUp(): void
+    {
+        $this->setupToken = false;
+        parent::setUp();
+
+    }
+
+
+    public function provider(): array
```
mshappe, line 19: \o/

mshappe (approved):

PR #515, opened 2021-03-02
Ensure database is updated before running tests

AricStewart:


tools/test_prereqs.php
```diff
@@ -13,6 +15,8 @@
 
 $db = MyPDO::instance();
 
+\db_do_update(0, SCHEMA::$REQUIED_DB_SCHEMA);
```
mshappe, line 18: This should already be covered by `create_schema.php`, though....

AricStewart: I was finding it was not... so I did not have any of the oauth2 tables when starting from after a nuke.

AricStewart: I think the update needs to be called explicitly.  Now maybe this should move to `create_schema.php`

mshappe: OK, now I see -- there's an older layer of updates that needs to run in addition to the newer modularized schema magic. I feel like they should be folded together, so I am going to experiment with moving the call into `DB::__construct`.

mshappe: Hm...I think rearranging this needs more thought. I don't want to just yank it around. I'm going to say your solution here is "good enough" for now, except that it should be in create_schema.php. To unstick myself for now, that's what I'm going to do on my parameterization branch. We can resolve this later.

mshappe (approved):

PR #516, opened 2021-03-02
Aric/reg tests

AricStewart:


api/src/Modules/registration/Controller/Ticket/BaseTicketInclude.php
```diff
@@ -0,0 +1,73 @@
...
+
+    }
+
+
+    public function processIncludes(Request $request, Response $response, $args, $values, &$data)
+    {
+        if (empty($values)) {
+            return;
+        }
+        if (in_array('ticketType', $values)) {
```
mshappe, line 28: This is fine for now, but we probably want a way to DRY this pattern up and move it to BaseController.

api/src/Tests/Cases/RegistrationTest.php
```diff
@@ -0,0 +1,148 @@
...
+
+
+    public function testTicket(): void
+    {
+        $this->runRequest('GET', '/registration/ticket/type/-1', null, null, 404);
+        $type = $this->runSuccessJsonRequest('GET', '/registration/ticket/type');
+        $data = $this->runSuccessJsonRequest('GET', '/registration/ticket/type/'.$type->id);
+        $this->assertEquals($type, $data);
+
+        $this->runRequest('POST', '/registration/ticket', null, null, 400);
```
mshappe, line 66: Could this block be done with a data provider?

api/src/Modules/registration/Controller/GetOpen.php
```diff
@@ -28,11 +28,19 @@ public function buildResource(Request $request, Response $response, $args): arra
             throw new NotFoundException('Event not found');
         }
 
-        $sql = "SELECT * FROM `Events` WHERE NOW() BETWEEN `dateFrom` AND `dateTo` AND `EventID` = $event";
+        $sql = "SELECT * FROM `Events` WHERE `EventID` = $event";
```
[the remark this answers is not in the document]
AricStewart: I want to be able to differentiate between  when a given EventID is invalid and when it is not during the events dates.
The first query did not let me do that.

api/src/Tests/Cases/RegistrationTest.php
AricStewart: I tried but we used $type->id and the data provider it actually run before any of the class is initialized.  So we cannot hav any figured or dynamic values like the id from a type we query from the DB.  We could partly dry it up by maybe proving all BUT that value and having some flag that replaces a placeholder with the value, but that started to get convoluted and I was unsure if it was cleaner or not.

api/src/Modules/registration/Controller/GetOpen.php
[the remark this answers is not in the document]
mshappe: That makes sense. Thank you.

api/src/Tests/Cases/RegistrationTest.php
mshappe: OK, that's legit. I'll add contemplating that kind of DRYing up to the pile of background-thinking I'm doing over how to better dance with PHPUnit.

mshappe (approved):

PR #517, opened 2021-03-03
Rename concom to staff in the API

AricStewart:
Cleanup some of the APIs and restore functionality


api/src/Modules/staff/Controller/ListStaff.php
```diff
@@ -0,0 +1,199 @@
...
+
+    private function listBuild($event)
+    {
+        global $Departments;
+
+        if ($event === null) {
+            return array();
+        }
+        $sql = <<<SQL
+    SELECT
```
mshappe, line 23: This query feels overcomplicated. It might not actually, but it feels like it. One reason I'm starting to push back on using raw SQL is it gets hard to follow what we're actually trying to do. So...what are we actually trying to do?

AricStewart: It is complicated and is directly pulled from `modules/concom/functions/LIST.inc`. I want to make another pass here and try to simplify a lot of this logic anyway but basically the whole issue is that the information we want is spread across 4 tables and we are trying to pull it all into 1 query. 

This was pulled from the old module and needs a rewrite to be more proper anyway but i was going to make that a next step

mshappe: Sounds good. My first thought is that this is a job for Join rather than subqueries, but that's just a first impression. I agree it can wait, one way or the other.

mshappe (approved):

PR #519, opened 2021-03-05
Aric/oas3

AricStewart:
Implementation of the first resource.

Knowledgeable comments really desired!

### DIFF
--- a/api/src/Controller/Announcement/ListDepartmentAnnouncements.php
+++ b/api/src/Controller/Announcement/ListDepartmentAnnouncements.php
@@ -3,6 +3,63 @@
     require_module 'standard';
 .*/
 
+/**
+ *  @OA\Get(
+ *      tags={"departments"},
+ *      path="/department/{id}/announcements",
+ *      summary="Lists announcements for a given department",
+ *      @OA\Parameter(
+ *          description="The id or name of the department",
+ *          in="path",
+ *          name="id",
+ *          required=true,
+ *          @OA\Schema(
+ *              oneOf = {
+ *                  @OA\Schema(
+ *                      description="Department name",
+ *                      type="integer"
+ *                  ),
+ *                  @OA\Schema(
+ *                      description="Department id",
+ *                      type="string"
+ *                  )
+ *              }
+ *          )
+ *      ),
+ *      @OA\Parameter(
+ *          description="Include the resource instead of the ID.",
+ *          in="query",
+ *          name="include",
+ *          required=false,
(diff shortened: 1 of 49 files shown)


.htaccess
```diff
@@ -53,3 +53,5 @@ Options -Indexes
     php_value session.cookie_lifetime 3600
     php_value session.gc_maxlifetime 3600
 </IfModule>
+
+Header set Access-Control-Allow-Origin "*"
```
mshappe, line 57: At some point, we're going to have to think about this one, because it has some implications for security.

mshappe (approved):

.htaccess
AricStewart: Yeah this piece is NOT required except to get swagger.io to work with the docker branch.
All the CORS stuff will NOT be included in the actaul pull request.

mshappe:
Can we just make an exception for swagger.io (I know, this requires more thinking about CORS, probably -- my own understanding of CORS is a bit of a jumble)

AricStewart:
Now with a built in swagger front end for development and testing:

http://localhost:2080/swagger

api/src/Controller/Announcement/ListDepartmentAnnouncements.php
```diff
@@ -3,6 +3,63 @@
...
+ *          in="query",
+ *          name="include",
+ *          required=false,
+ *          explode=false,
+ *          style="form",
+ *          @OA\Schema(
+ *              type="array",
+ *              @OA\Items(
+ *                  type="string",
+ *                  enum={"departmentId","postedBy"}
```
mshappe, line 40: `department` rather than `departmentId`, but that can also wait until we do the shift to snake_case, since we discussed it all in a bundle.

api/src/Controller/Event/PostEvent.php
```diff
@@ -3,6 +3,51 @@
...
 .*/
 
+/**
+ *  @OA\Post(
+ *      tags={"events"},
+ *      path="/event",
+ *      summary="Adds a new event",
+ *      @OA\RequestBody(
+ *          @OA\MediaType(
+ *              mediaType="multipart/form-data",
```
mshappe, line 13: We should decide if, since we're publishing this now (not that anyone but us is using it yet), if we need to version things when we start shifting these all to JSON (and start shifting the parameter cases as we discussed)

mshappe (approved):
Love where this is going so far!

PR #520, opened 2021-03-06
Rework how we process API includes

AricStewart:
less code duplication and more directly understandable

### DIFF
--- a/api/src/Controller/BaseController.php
+++ b/api/src/Controller/BaseController.php
@@ -38,6 +38,53 @@ class ConflictException extends Exception
 
 }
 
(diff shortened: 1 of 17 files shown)


api/src/Controller/BaseController.php
```diff
@@ -38,6 +38,53 @@ class ConflictException extends Exception
 
 }
 
+class IncludeResource
```
mshappe, line 41: Not a stopper for this PR, but ideally we should be one class per file. It's not really a big deal, though.

mshappe (approved):

PR #524, opened 2021-03-19
Aric/api cleanup

AricStewart:


api/src/Controller/Cycle/GetCycle.php
```diff
@@ -43,21 +43,15 @@ class GetCycle extends BaseCycle
...
-        $cycles = $sth->fetchAll();
-        if (empty($cycles)) {
-            throw new NotFoundException('Cycle Not Found');
-        }
-        $this->id = $args['id'];
-        $entry = $cycles[0];
-        $entry['id'] = $this->id;
-        unset($entry['AnnualCycleID']);
+        $cycle = $this->getCycle($params)[0];
+        $cycle['id'] = $cycle['AnnualCycleID'];
```
mshappe, line 49: Just as a suggestion -- and this is something I think I took advantage of elsewhere -- you can handle this renaming in the SELECT statement. Instead of `SELECT *`, select the specific fields with aliases, e.g. `SELECT AnnualCycleID as id`, etc. I can do this even with Atlas\Query\Select, so the pattern would hold either way. For GETs, this also seems in general like the best way to handle our name-translation issues.

mshappe (approved):

api/src/Controller/Cycle/GetCycle.php
AricStewart: That is a great second pass cleanup idea. I will put it on my note.

PR #528, opened 2021-03-26
API Clean-up SQL in staff module to use Atlas

AricStewart:


api/src/Modules/staff/Controller/BaseStaff.php
```diff
@@ -197,5 +140,75 @@ protected function buildEntry(Request $request, $id, $dept, $member, $note, $pos
...
 
+    protected function selectStaff($event = null, $department = null, $member = null)
+    {
+        if ($event === null) {
+            $event = $this->getEvent('current')['id'];
+        }
+
+        $select = Select::new($this->container->db);
+
+        $historical = $select->subselect()->columns(
```
mshappe, line 151: So, this gets into both a general philosophy-of-formatting thing, and an actual "our linter is wrong and I don't care what it believes" thing.

The format I prefer looks something like this:

```
$select->subselect()->columns(...)
      ->from(...)
      ->where(...)
```

but our linter currently chokes on the free-standing arrows. We need to fix that, because the linter is objectively wrong :grin:

The format you're using here is VERY readable...if one has a high enough resolution screen that it all fits on the vertical. On even a 1920x1080 screen I'd need to scroll 3-4 times to see the whole query (same with the SQL for that matter) and I'd lose track of it.

I don't want to get to deep into the weeds on this one, though--we've spent enough time bikeshedding.

mshappe (approved):
Other than my formatting bikeshedding, this looks good!

PR #537, opened 2021-04-04
Aric/controller update

AricStewart:


api/src/Controller/Announcement/ListMemberAnnouncements.php
```diff
@@ -66,66 +67,30 @@ public function buildResource(Request $request, Response $response, $args): arra
...
-                )
-            ))
-            ORDER BY `PostedOn` ASC
-SQL
-        );
-        $sth->execute();
-        $todos = $sth->fetchAll();
+        $select = Select::new($this->container->db);
+
+        $sub1 = $select->subselect()->columns('COUNT(AccountID)')->from('ConComList')->whereEquals(['AccountID' => $user]);
```
mshappe, line 72: This is where I want us to fix our phpcs rules (once I figure out exactly how) to allow hanging arrows. There has to be a way. I'll find it eventually.

api/src/Controller/BaseController.php
```diff
@@ -454,53 +455,19 @@ public function arrayResponse(Request $request, Response $response, $data, $code
...
-                WHERE
-                    DepartmentID = d1.DepartmentID
-            ) AS email
-            FROM
-                Departments d1
-            WHERE
-SQL;
+        $select = Select::new($this->container->db);
+        $select->columns('*');
+        $select->columns($select->subselect()->columns('COUNT(DepartmentID)')->from('`Departments` d2')->whereEquals(['d2.ParentDepartmentID' => 'd1.DepartmentID'])->andWhere("NOT NAME = 'Historical Placeholder'")->as('childCount')->getStatement());
```
mshappe, line 460: Including the alias in the from arguments is fine IMO, but you could also say, I believe, `->from('Departments')->as('d2')`. This is not a change you need to make here. This is just informational.

mshappe (approved):

PR #541, opened 2021-04-06
Add a few missing OAS blocks

AricStewart:
And correct a few others

### DIFF
--- a/ciab.openapi.yaml
+++ b/ciab.openapi.yaml
@@ -201,9 +201,9 @@ paths:
           multipart/form-data:
             schema:
               properties:
-                Text:
+                text:
                   type: string
-                Scope:
+                scope:
                   type: integer
               type: object
       responses:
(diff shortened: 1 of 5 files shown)


ciab.openapi.yaml
```diff
@@ -750,13 +750,13 @@ paths:
           multipart/form-data:
             schema:
               properties:
-                From:
+                dateFrom:
```
mshappe, line 753: So, did we settle on camelCase? For some reason, my brain had it that we'd settled on snake_case (and that's what I used in my store update the other day)...

mshappe (approved):

ciab.openapi.yaml
AricStewart: My brain has that we had not settled yet... :)  I was just making everything match for now so that when we decided it would be eaiser to come in and fix it.

So snake_case is it?

PR #542, opened 2021-04-07
Aric/snake case

AricStewart:
This is all parts of the API except for members, which is a bit of a mess still and needs some descrambling anyway.


mshappe (approved):
And now we sort into house Slytherin, as it should be :-D

PR #544, opened 2021-04-10
Implement getMember to replace \lookup_xxx and findMember calls

AricStewart:
This does an important thing with an important side effect:

    * Continues to disconnect us from the legacy PHP system

However also removes API level Neon support.

The removal of Neon support should be a non-issue because the legacy
web code should be importing all members into the database as they are
created.  But it is important to note that changes made to members via
the API will not propigate back into Neon either.


AricStewart:
This should be the last snake_case major change to the API.. I think we should be able to be stable after this.

mshappe:
I guess I had been a bit fuzzy as to whether the sync was currently two-way, anyway. Is there anything we're relying on in that two-way sync that we'll be sorry to lose?

api/src/Controller/Member/FindMembers.php
```diff
@@ -44,28 +66,118 @@
...
+        if ($from == 'all') {
+            $from = "email,id,legal_name,name,badge";
+        }
+        $from = explode(',', trim($from));
+        $p = $request->getQueryParam('partial', 'false');
+        $partial = filter_var($p, FILTER_VALIDATE_BOOLEAN);
+
+        $select = Select::new($this->container->db)
+            ->columns(...BaseMember::selectMapping())
+            ->columns('(CASE WHEN `PreferredFirstName` IS NOT NULL THEN `PreferredFirstName` ELSE `FirstName` END) AS `first_name`')
```
mshappe, line 99: This bit with the case/when is nicely done!

api/src/Controller/Member/PostMember.php
```diff
@@ -51,53 +52,47 @@ class PostMember extends BaseMember
...
+            throw new InvalidParameterException("Required 'legal_first_name' and/or 'legal_last_name' parameter not present");
         }
 
-        if (array_key_exists('Neon', $GLOBALS)) {
-            $accountID = \neon_createUser(
-                $body['email1'],
-                $body['firstName'],
-                $body['lastName']
-            );
+        $currentIdTop = Select::new($this->container->db)
```
mshappe, line 72: I assume this is here because `AccountID` is not currently a proper auto-increment primary key, but an import from Neon? I suppose as long as we're allowing for both cases, we're stuck with this. Practically speaking, it's probably safe right now, but long-term it isn't -- it isn't safe against concurrent access.

Probably we need to either A] add an auto-increment to this column or B] transition to a MemberID column that's the true primary key for the table, with the legacy AccountID being ONLY for Neon IDs. Thoughts?

mshappe (commented):
One big question mark but otherwise, this looks good.

api/src/Controller/Member/PostMember.php
AricStewart: No, the reason here is that there was expressed a desire that badge number should start at `1000` and never start at just `1`.  `AccountID` is auto-incrementing but I could not find an easy way in mysql to set a minimum value..

mshappe (approved):

PR #548, opened 2021-05-09
Bye bye, leftnav

mshappe:
New UX does not include "leftnav", which was misleading anyway (told you you had a badge number even if you weren't registered). Could have been fixed, but since the new UI doesn't include it, best to excise it now.


AricStewart:
Do we want to delete the file `leftnav.inc` as well?  If it is not included anywhere we may as well not have it.  Leaving dead code around make things messy.

mshappe:
> Do we want to delete the file `leftnav.inc` as well? If it is not included anywhere we may as well not have it. Leaving dead code around make things messy.

Yeah, I was right on the fence about it. Just pushed a commit whacking it.

PR #550, opened 2021-05-17
Phase 1 Check in and Boarding Revamp

mshappe:
* Simplify markup/classes to maximize real estate on small screens. NOTE: this needs a second pass
* Separate Checkin/Boarding/Lost into three pages. NOTE: navigation needs another pass
* Convert a bunch of DOM-manipulation markup to template strings for easier reading/maintenance
* Add some capabilities to ListTickets

As stressed above, a second pass is still needed to really make this pretty and improve navigation, but this is a few inches closer to the wireframes.


mshappe:
(As currently planned, second pass will kick off this Sunday)

AricStewart (approved):

PR #551, opened 2021-05-22
Add Vendor extensions for the API

AricStewart:
The first extenstion is the beginning of a Stripe payment processing
system.

### DIFF
--- a/api/src/Controller/BaseVendorController.php
+++ b/api/src/Controller/BaseVendorController.php
@@ -1,0 +1,237 @@
+<?php declare(strict_types=1);
+/*.
+    require_module 'standard';
+.*/
+
+namespace App\Controller;
+
+use Slim\Container;
+use Slim\Http\Response;
+use Slim\Http\Request;
+
+abstract class BaseVendorController extends BaseController
+{
+
+    protected static $instance;
+
+
+    protected function __construct(string $api_type, Container $container)
+    {
+        $basename = explode('\\', get_class($this));
+        if (strcasecmp($basename[0], 'App') == 0) {
+            array_shift($basename);
+            array_shift($basename);
+            array_unshift($basename, 'Vendor');
+            $vendor = implode('\\', $basename);
+            if (class_exists($vendor)) {
+                self::$instance = new $vendor($container);
+            }
+        }
+        parent::__construct($api_type, $container);
+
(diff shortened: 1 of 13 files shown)


api/src/Controller/BaseVendorController.php
```diff
@@ -0,0 +1,237 @@
...
+namespace App\Controller;
+
+use Slim\Container;
+use Slim\Http\Response;
+use Slim\Http\Request;
+
+abstract class BaseVendorController extends BaseController
+{
+
+    static protected $instance;
```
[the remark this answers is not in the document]
mshappe: @AricStewart Take note ^^. I've noted this inversion a couple of times.

api/src/Vendors/Stripe/PostPayment.php
```diff
@@ -0,0 +1,48 @@
...
+
+
+    public function buildResource(Request $request, Response $response, $params): array
+    {
+        $required = ['success', 'cancel', 'price', 'quantity', 'item'];
+        $body = $this->checkRequiredBody($request, $required);
+
+        $checkout_session = \Stripe\Checkout\Session::create([
+            'payment_method_types' => ['card'],
+            'line_items' => [[
```
mshappe, line 20: So, right now, this appears to be fixed at a single line item; we will want to iterate over the line items in a cart.

mshappe: This also means that this endpoint could actually just take a shopping_cart_id rather than take the details as part of the POST, if I understand the workflow.

mshappe (dismissed):
A good start!

api/src/Tests/Cases/PaymentTest.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+
+    public function testPayment(): void
+    {
+        global $BASEURL, $_ENV;
+
+        if (array_key_exists('STRIPE_PRIVATE_KEY', $_ENV)) {
+            $this->runRequest(
+                'POST',
```
mshappe, line 36: Can these arguments be put in an Array and then splatted into arguments, so they don't have to be copypasta'd?

mshappe (approved):
Just some DRYcleaning in the test.

PR #554, opened 2021-06-04
Add API interface to list tickets printed but not picked up

AricStewart:


mshappe:
...did I not already submit a PR that contained some of this concept? I could swear I added at least some of this to the API...

mshappe:
> ...did I not already submit a PR that contained some of this concept? I could swear I added at least some of this to the API...

Never mind -- I did not read to see WHICH endpoint this was!

mshappe (approved):

PR #558, opened 2021-06-23
Escape member data when modifying concom membership

AricStewart:
Fixes the problem with special characters like ' appearing in a name


mshappe (approved):
That'd do it. It occurred to me to wonder if this link really needed all of this information in it, but that's basically an API change and we don't need to do that right this minute.

PR #560, opened 2021-06-24
API: Sanitize &#xx; codes coming out of the DB

AricStewart:


api/src/Controller/BaseController.php
```diff
@@ -665,7 +665,11 @@ public static function insertPayloadFromParams(array $params, $includeId = true)
         };
 
         foreach ($params as $key => $val) {
-            $ret[$paramsToColumns[$key]] = $val;
+            if (is_string($val)) {
+                $ret[$paramsToColumns[$key]] = htmlspecialchars_decode($val, ENT_QUOTES);
```
mshappe, line 669: AH-HAH!

mshappe (approved):

PR #566, opened 2021-07-13
Aric/event to api

AricStewart:


api/src/Controller/Event/PutEvent.php
```diff
@@ -0,0 +1,130 @@
...
+        $permissions = ['api.put.event'];
+        $this->checkPermissions($permissions);
+
+        $event = $this->getEvent($params['id']);
+        $body = $this->checkRequiredBody($request, []);
+
+        unset($body['cycle']);
+        $cycle_from = $event['cycle'];
+        $cycle_to = $event['cycle'];
+        if (array_key_exists('date_from', $body)) {
```
mshappe, line 77: These inner blocks feel like they could be factored out into a method.

mshappe (dismissed):
Just some DRY advice, and otherwise, this looks good!

AricStewart:
Good points... Reworked.

api/src/Controller/Event/BaseEvent.php
```diff
@@ -117,5 +118,59 @@ public function __construct(Container $container)
...
+
+        if ($event) {
+            $cycle_from = $event['cycle'];
+            $cycle_to = $event['cycle'];
+        } else {
+            $cycle_from = null;
+            $cycle_to = null;
+        }
+        if (array_key_exists('date_from', $body)) {
+            try {
```
mshappe, line 134: Um...not actually seeing any difference to this code...

AricStewart: Oh I missed your point. I thought you meant the blocks between the PUT and POST methods which I did combine.  Not the methods WITHIN the function.  I can take another pass later today if you feel it worth it.

mshappe: Oh, yeah, sorry. Refactor all of that inner logic for handling date to/from to a method, please :-D

AricStewart: Ok fixed.

mshappe (approved):
That's the stuff!

PR #578, opened 2022-01-18
Aric/style

AricStewart:


mshappe (approved):
ESlint still seems to be unhappy tho'

PR #585, opened 2022-02-19
Aric/vol status

AricStewart:


AricStewart:
the eslint test always fails when a source module is removed.  It is a false error.

api/src/Tests/Cases/VolunteersTest.php
```diff
@@ -126,13 +126,26 @@ public function testHours(): void
...
 
         /* Overlap Testing */
-        $base = date('Y-m-d h:i:s', strtotime('+1 Year'));
-        $plusMin = date('Y-m-d h:i:s', strtotime('+1 Year +30 minutes'));
-        $minusMin = date('Y-m-d h:i:s', strtotime('+1 Year -30 minutes'));
-        $plusHour = date('Y-m-d h:i:s', strtotime('+1 Year +1 hour'));
-        $minusHour = date('Y-m-d h:i:s', strtotime('+1 Year -2 hours'));
-        $largeEnd = date('Y-m-d h:i:s', strtotime('+1 Year +4 hours'));
-        $subEnd = date('Y-m-d h:i:s', strtotime('+1 Year -30 minutes'));
+        $date = new \DateTime('+1 Year');
```
mshappe, line 129: Since the date intervals are specified as strings, could this be an array of date intervals and a loop?

modules/concom/report/sitesupport/report.js
```diff
@@ -0,0 +1,10 @@
+/* jshint esversion: 6 */
+/* globals Vue */
+
+import reportGeneration from '../../../../sitesupport/reportGeneration.vue.js'
```
mshappe, line 4: It feels like these could be DRY'd up by moving them up to general sitesupport, but I also know we're trying to keep things modular, so maybe not?

modules/volunteers/report/sitesupport/volunteerReport.js
```diff
@@ -0,0 +1,97 @@
...
+
+export default {
+  props: {
+    reportDisplay: String,
+    haveEvent: Boolean,
+    header: String
+  },
+  created() {
+    if (this.haveEvent) {
+      apiRequest('GET', '/event/current', null)
```
mshappe, line 12: There's a race condition here--these could still return out of sequence, which would cause `gotEvent` to do the wrong thing, I think. I might be wrong about this in practice, but you might want to think about making this more of a then-chain, e.g.

```
apiRequest(...'/event/current'...)
  .then(...)
  .then(-> apiRequest(...'/event'...)
  .then(...)
```

I'm not 100% on the syntax, but something like that.

modules/volunteers/sitesupport/volunteersEntry.js
```diff
@@ -0,0 +1,292 @@
...
+
+/* jshint esversion: 6 */
+/* globals apiRequest, Vue, userLookup */
+/* exported root */
+
+var volApp = Vue.createApp({
+  created() {
+    console.log('load volunteer hours pane');
+    this.setEndDate();
+    apiRequest('GET', '/department','max_results=all')
```
mshappe, line 13: Here, again, if order matters, I'd consider making it a then-chain. But only if order matters.

modules/volunteers/sitesupport/volunteersEntry.js
```diff
@@ -0,0 +1,292 @@
...
+        }
+      }
+    },
+    onFail(target, resp, name, code) {
+      userLookup.markFailure();
+      document.getElementById('submit').disabled = true;
+      this.uid = false;
+      this.volunteerName = 'a Volunteer';
+
+      if (code == 404) {
```
mshappe, line 207: No handler for 418? :-D

(II had occasion this week to introduce my European padawan to RFC2324)

mshappe (dismissed):
Check out those possible race conditions, but otherwise, this looks good! I'm glad you're taking to Vue so well! It's pretty nifty, I think.

api/src/Tests/Cases/VolunteersTest.php
AricStewart: I know. It is strangely picky.

modules/concom/report/sitesupport/report.js
AricStewart: I renamed these files to vue.js because they will be likely used for other vue components as we move forward anyway.
I am for keeping things modular for now, we can DRY them up later if fitting.

modules/volunteers/report/sitesupport/volunteerReport.js
AricStewart: I don't think so. or at least I am not seeing a problem. 
gotCurrentEvent just sets this.selectedEvent, which if it comes in 1st will not do any hard and if it comes in 2nd will set the select box.
gotEvent populates that listbox and if selectedEvent is set makes sure the right entry is selected.

modules/volunteers/sitesupport/volunteersEntry.js
AricStewart: Again I tried to make sure it didnt.

mshappe (approved):

PR #599, opened 2023-07-16
Fix configure_system script and Cleanup Documentation

rkaiser-dev:
* Removed duplicated documentation and consolidated into `CONTRIBUTING.md`
* Re-organize other documentation into `doc/`
* Fix broken function used by `configure_system.php`


doc/CONTRIBUTING.md
AricStewart, line 1: Just curious why the combination of doc files all into here?

doc/RELEASE.md
AricStewart, line 1: Good Catch!

functions/users.inc
```diff
@@ -514,19 +514,18 @@ function createPassword(/*.string.*/$id, /*.string.*/$password): string
         $duration = '+1 year';
     }
 
-    $value = lookup_user_by_login($id);
+    $value = lookup_users_by_email($id);
```
AricStewart, line 517: Are you sure this is what we want to do? 
Not using `lookup_users_by_login`?
I see the very first call in `lookup_users_by_login` is `lookup_users_by_email` but it also then fallsback to the CRM login field if it exists.

package-lock.json
AricStewart, line 1: Don't include this file in a diff unless you are explicitly adding/removing or upgrading a package.

AricStewart (changes requested):

doc/CONTRIBUTING.md
rkaiser-dev: To me it felt cleaner to have important things organized into a single directory other than project root. I would be open to leaving all of the files I moved into here in the project root instead, so long as setup/getting started info was contained in a single file. 

During setup I looked through each of the 3 documents that I ended up removing, and some of them contained information that was no longer relevant, or would have been more relevant while getting things set up for the first time.

functions/users.inc
rkaiser-dev: Interesting, when this was `lookup_users_by_login`, the Docker console output complained that the function was undefined. Perhaps it was another part of this that had been failing, I will double check.

I do know that changing to `lookup_users_by_email` did result in successfully completing the POST to `configure_system`. I'll check into this and report back.

package-lock.json
rkaiser-dev: I can remove the changes, however this is the resulting file after I used `npm install` with node version 18.x.

functions/users.inc
rkaiser-dev: I can see why it had been failing before, as well. `lookup_user_by_login` was missing the `s` in `users`, so that probably explains the failure.

rkaiser-dev: Also, I'm curious, but it seems like this function is only used by `configure_system` and not anywhere else. Is there a reason this might be the case?

rkaiser-dev:
Closing this PR as I found a few other things that I want to address. Will just start from a fresh branch.

PR #605, opened 2023-07-25
Convert Volunteers Admin to Vue

AricStewart:
There is a lot of work here and not all of these patches are 100% atomic clean but it is close and at least tracks the changes in a finer grain of detail than a single monolithic patch.


modules/volunteers/admin/sitesupport/editHours.js
```diff
@@ -0,0 +1,116 @@
...
+  methods: {
+    commitHours() {
+      confirmbox(
+        'Please! double check entries',
+        'Proceed with Volunteer Hour Update?').then(() => {
+        var item = {};
+
+        item['hours'] = parseFloat(this.record.hours);
+        item['end'] = this.record.end;
+        var e = document.getElementById('edit_mod');
```
rkaiser-dev, line 34: I can't quite put my finger on exactly why but it seems weird that we would find a need to use `document.getElementById` while using Vue. We should be able to bind to an `<option>` tag without issue and pull the value directly without the need to do this.

modules/volunteers/admin/sitesupport/editHours.js
```diff
@@ -0,0 +1,116 @@
...
+        end: null,
+      },
+    }
+  },
+  methods: {
+    commitHours() {
+      confirmbox(
+        'Please! double check entries',
+        'Proceed with Volunteer Hour Update?').then(() => {
+        var item = {};
```
rkaiser-dev, line 30: Another option for filling in object details would be like this:

```
const item = {
  hours: parseFloat(this.record.hours),
  end: this.record.end,
  // ignoring modifier for now until I have a chance to look at binding with `<option>` tags
  department: this.record.department.id,
  authorizer: this.record.authorizer.id
};
```

There's not much of a difference in how we decide to do it, but it can get a bit cleaner when it comes to things like arrays and nested objects if we need to add those.

sitesupport/common.js
```diff
@@ -22,6 +23,12 @@ function escapeHtml(text) {
   return text.replace(/[&<>"']/g, function(m) { return map[m]; });
 }
 
+function simpleObjectToRequest(obj) {
```
rkaiser-dev, line 26: This function may be why you are seeing weird behavior where `null` becomes `"null"` in your query params. I believe JS has some built-in tooling for handling URL encoded parameters in forms that we might consider leveraging.

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

rkaiser-dev: Even still, I believe it's the case that passing in an object with `{ foo: null }` will still result in `foo=null` here. So to support query params we may need to approach it with the magic values.

modules/volunteers/admin/sitesupport/checkout.js
```diff
@@ -0,0 +1,223 @@
...
+  },
+  methods: {
+    processCheckout() {
+      confirmbox('Confirm Distribute Gifts',
+        'Are the selected gifts correct?').then(() => {
+        showSpinner();
+        var prizes = [this.checkout.length, 0, 0];
+        this.checkout.forEach((item) => {
+          console.log(item);
+          for (let n = 0; n < item.count; n++) {
```
rkaiser-dev, line 21: How high of a count can we expect for items? If we have 50 items, this could potentially send 50 simultaneous requests and these will execute as promises and are non-blocking. If we want to block and wait for each one, or even limit to a chunk of requests at a time we may want to consider a different approach.

modules/volunteers/admin/sitesupport/checkout.js
```diff
@@ -0,0 +1,223 @@
...
+export default {
+  data() {
+    return {
+      checkout: [],
+      hoursSpent: 0,
+      groupsNow: [],
+      failedList: null,
+    }
+  },
+  methods: {
```
rkaiser-dev, line 13: It might make the code a bit easier to read if we extract some of these functions to another spot in the file, and then reference those functions in our `methods` object.

For example, you might have something like:

```
const processCheckout = () => {
  // Code to complete the checkout process
}

export default {
  data() {
    return {
       checkout: [],
       hoursSpent: 0,
       groupsNow: [],
       failedList: undefined // recommend undefined here instead of null, as null can have weird behaviors in JS and undefined is more typical unless Vue specifically requires null
    },
    methods: {
       processCheckout
    }
  }
}
```

modules/volunteers/admin/sitesupport/checkout.js
```diff
@@ -0,0 +1,223 @@
...
+    printHoursLeft() {
+      if (this.$parent.$refs.gfttbl) {
+        return this.$parent.printHours(this.$parent.$refs.gfttbl.hoursRemaining - this.hoursSpent);
+      }
+    },
+    printHoursSpent() {
+      return this.$parent.printHours(this.hoursSpent);
+    }
+  },
+  template: `
```
rkaiser-dev, line 165: I noticed Vue documentation provides another way to describe templates as well, where it looks more like HTML rather than a string. Is there a reason that we are using this approach that I haven't seen yet in Vue docs?

modules/volunteers/admin/sitesupport/checkout.js
rkaiser-dev: Actually, on seeing `failedList` being used below I'd recommend simply setting to `[]` instead as both the default and for cases where no items have failed to be checked out.

modules/volunteers/admin/sitesupport/editHours.js
rkaiser-dev: I found it - we can indeed bind to `<select>` which is what I was thinking and meant, rather than `<option>`.

According to docs it looks like we should be able to do something like:

- Define in `data`
`hour_modifier: 1`

In HTML:
```
<select class="UI-select" style="width:auto" name="TimeModifier" v-model="hour_modifier">
            <option value=0.5>Half Time - 1 hour = 0.5 hours credit</option>
            <option value=1>Normal - 1 hour = 1 hour</option>
            <option value=1.5>Time Plus Half - 1 hour = 1.5 hours credit</option>
            <option value=2>Double - 1 hour = 2 hours credit</option>
</select>
```

This should result in "Normal - 1 hour = 1 hour" being the option selected by default when the page loads. Changing the selected option should be accessible via `hour_modifier` and when we go to reference this when building `item` for our request with `this.record.hour_modifier`.

https://vuejs.org/guide/essentials/forms.html#select

modules/volunteers/admin/sitesupport/editHours.js
```diff
@@ -0,0 +1,116 @@
...
+      confirmbox('DELETE Volunteer Entry?').then(() => {
+        apiRequest('DELETE', '/volunteer/hours/' + this.record.id)
+          .then(() => {
+            location.reload();
+          });
+      });
+    },
+    show(record) {
+      /* deep copy */
+      this.record = JSON.parse(JSON.stringify(record));
```
rkaiser-dev, line 55: Can you help me understand why we need to deep copy `record` here?

modules/volunteers/admin/sitesupport/processReturn.js
```diff
@@ -0,0 +1,89 @@
...
+      this.claims.forEach((c) => {
+        if (c.returned && c.reward.promo != '1') {
+          total += parseFloat(c.reward.value);
+        }
+      });
+
+      this.creditHours = Math.round(total * 100) / 100;
+    },
+    show(claims) {
+      this.claims = JSON.parse(JSON.stringify(claims));
```
rkaiser-dev, line 41: Similar question, can you help me understand the need to deep copy here?

modules/volunteers/admin/sitesupport/editHours.js
AricStewart: The deep copy is required so that we do not edit the record passed to us from the table itself. Otherwise as we change values the change live on that record and we would require robust undo logic to resort the state if the edit sidebar is closed or canceled.

modules/volunteers/admin/sitesupport/processReturn.js
AricStewart: The deep copy is required so that we do not edit the record passed to us from the table itself. Otherwise as we change values the change live on that record and we would require robust undo logic to resort the state if the edit sidebar is closed or canceled.

modules/volunteers/admin/sitesupport/editHours.js
rkaiser-dev: When you say "from the table itself", do you mean the table component defined in `hourTable` or do you mean like a database table record?

modules/volunteers/admin/sitesupport/checkout.js
AricStewart: I believe it is called `object-based syntax`, but it is also referred to as non-script setup SFC format.  The basic reason is that I was initial taught Vue using it and it has done everything I need so I have not moved on from there.

modules/volunteers/admin/sitesupport/editHours.js
AricStewart: Yes, from the hourTable.

modules/volunteers/admin/sitesupport/editHours.js
AricStewart: Agreed, this is a place I did not clean up as much as I should, I will do a second pass. I could just drop the english wording "Double, Time Plus Half, etc.." then it because a very easy thing to populate.

modules/volunteers/admin/sitesupport/editHours.js
AricStewart: oh nice, I was unaware you could have code in your initializer like that.  That is neat!

sitesupport/common.js
AricStewart: Yeah, for now documented magic values, Tomorrow the world.

modules/volunteers/admin/sitesupport/checkout.js
```diff
@@ -0,0 +1,223 @@
...
+                } else {
+                  item.failure = response.responseText;
+                }
+                prizes[2] += 1;
+                if (this.failedList == null) {
+                  this.failedList = [ item ];
+                } else {
+                  this.failedList.append(item);
+                }
+                const target = this.checkout.findIndex(element => element.id == item.id);
```
rkaiser-dev, line 43: For JS it is typically best practice to use `===` when dealing with equality.

modules/volunteers/admin/sitesupport/editHours.js
rkaiser-dev: And to make sure I understand - we _don't_ want to track those changes if the sidebar has closed? Is that because this component itself has been setup to post its own data to the server?

modules/volunteers/admin/sitesupport/checkout.js
AricStewart: I would be surprised at 50, I would say at most 10.  We sort of block and wait as we have a please wait spinner going until all the request have succeeded or failed.  Then we report the result.  I do not feel this is terrible (but correct me if it make your web developer skin crawl)  we could modify the API endpoint to accept a list and then report results for each item.  Then it would be 1 request.

modules/volunteers/admin/sitesupport/checkout.js
AricStewart: Makes sense I will look and dig.

modules/volunteers/admin/sitesupport/editHours.js
rkaiser-dev: I think having the text in the `<option>` is fine, the value will be what gets sent to the back-end anyway. Whatever will be easiest for folks to understand is ideal, here.

modules/volunteers/admin/sitesupport/checkout.js
rkaiser-dev: Depends on the implementation - my concern is mostly, do we care if 10 requests are fire simultaneously? It is only 1 user click, but 10 requests all at once. 

Sending the full list of items and processing the full list would be ideal here if we can do that.

modules/volunteers/admin/sitesupport/checkout.js
rkaiser-dev: I see it now, thanks for pointing that out. Looks like this is the approach if we are not doing build-time things to it, so that makes sense.

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -0,0 +1,297 @@
...
+  },
+  beforeUpdate() {
+    this.groupIndex = -1;
+    this.groupColorId = null;
+    this.lastGroup = null;
+    this.hoursRemaining = this.totalHours - this.hoursSpent;
+  },
+  methods: {
+    clicked(record) {
+      if (this.$parent.userId == null && this.$parent.isAdmin) {
```
rkaiser-dev, line 147: What is `$parent` here? If we're using this to refer to other components, could we take advantage of component registration instead?

https://vuejs.org/guide/components/registration.html#local-registration

modules/volunteers/admin/sitesupport/editHours.js
```diff
@@ -0,0 +1,116 @@
...
+        },
+        department: {
+          name: '',
+        },
+        end: null,
+      },
+    }
+  },
+  methods: {
+    commitHours() {
```
rkaiser-dev, line 26: Not sure if we're able to do so within the components, but I think we can - it might be nice to prefer `async / await` here rather than `promise.then()` syntax. It's a bit cleaner and allows us to do `try/catch` instead of `promise.then().catch()` and can be a bit cleaner code wise.

An example of how this would look is like this

```
methods: {
  async commitHours() {
     // Code before the API request

    await apiRequest('PUT', '/volunteer/hours/' + this.record.id + '?force=1', simpleObjectToRequest(item));
    location.reload();
  }
}
```

In more complex scenarios you could assign the result of the request to a variable like anything else:

`const response = await apiRequest()`

rkaiser-dev: Could definitely be a "as we clean up the code" thing - no rush on this of course.

modules/volunteers/admin/sitesupport/editHours.js
AricStewart: Correct

modules/volunteers/admin/sitesupport/giftTable.js
AricStewart: the $parent is the parent component in the DOM.  in this case it is `modules/volunteers/admin/sitesupport/vue.js` that holds all the components of the admin page.

modules/volunteers/admin/sitesupport/editHours.js
AricStewart: I am unfamiliar with async/await so I would love to learn more about them.  I say we stick with promise then for now but if there is a good way to  open up async/await as a option that would be great!

rkaiser-dev (approved):

PR #608, opened 2023-07-31
Aric/volunteers retire

AricStewart:


modules/volunteers/admin/sitesupport/editPrize.js
```diff
@@ -45,11 +45,20 @@ export default {
       showSidebar('edit_prize_div');
     },
     deletePrize() {
-      confirmbox('DELETE Gift Entry?',
-        'WARNING!!!<br>  Only do this if NONE of this gift ' +
-        'has been distributed. <br>It will lead to corrupt ' +
-        'reward records. <br>To DELETE a gift that has been ' +
-        'rewarded set inventory to \'0\'').then(() => {
+      var warning = `
```
rkaiser-dev, line 48: Instead of `var`, can we use `let` instead for `warning` and `title`?

modules/volunteers/admin/sitesupport/hourEntryField.js
```diff
@@ -0,0 +1,74 @@
...
+      },
+      set(value) {
+        this.$emit('update:modelValue', value);
+        this.$emit('change');
+      }
+    },
+    hours: {
+      get() {
+        if (this.enterHours != null) {
+          var v = this.enterHours;
```
rkaiser-dev, line 42: It looks like we're not reassigning the value at all here, so can we use `const` instead of `var` for it?

modules/volunteers/admin/sitesupport/hourEntryField.js
```diff
@@ -0,0 +1,74 @@
...
+        return Math.floor(this.modelValue);
+      },
+      set(value) {
+        this.enterHours = Math.floor(value);
+      }
+    },
+    mins: {
+      get() {
+        if (this.enterMin != null) {
+          var v = this.enterMin;
```
rkaiser-dev, line 55: It looks like we're not reassigning the value at all here, so can we use `const` instead of `var` for it?

rkaiser-dev (approved):
Feel free to take or leave the choice to switch from `var` to `let` and `const`. In my experience, it seems to be recommended to avoid `var` and prefer the other options instead.

PR #609, opened 2023-08-07
Volunteer improvements

AricStewart:
Reduce full page reloads
Clearly display selected volunteers name
Shift a few more things to vue and out of raw javascript.

### DIFF
--- a/modules/volunteers/admin/sitesupport/giftTable.js
+++ b/modules/volunteers/admin/sitesupport/giftTable.js
@@ -1,6 +1,35 @@
 /* jshint esversion: 6 */
 /* globals apiRequest */
 
+const COLUMNSBASE = [
+  {value:'name', title:'Name '},
+  {value:'promo', title:'Promotional'},
+  {value:'inventory', title:'Total Inventory'},
+  {value:'value', title:'Value '},
+  {value:'remaining', title:'Remaining'},
+  {value:'limit', title:'Limit'}
+];
+
+const COLUMNSUSER = [
+  {value:'name', title:'Name '},
+  {value:'promo', title:'Promotional'},
+  {value:'inventory', title:'Total Inventory'},
+  {value:'value', title:'Value '},
+  {value:'remaining', title:'Remaining'},
+  {value:'limit', title:'Limit'},
+  {value:'acquired', title:'Acquired'}
+];
+
+const GROUPSTYLES = [
+  'VOL-color-amber', 'VOL-color-aqua', 'VOL-color-brown',
+  'VOL-color-cyan', 'VOL-color-indigo', 'VOL-color-khaki',
+  'VOL-color-lime', 'VOL-color-orange', 'VOL-color-pink',
+  'VOL-color-purple', 'VOL-color-red', 'VOL-color-sand',
+  'VOL-color-teal', 'VOL-color-yellow', 'VOL-color-deep-purple',
+  'VOL-color-deep-orange', 'VOL-color-light-blue',
+  'VOL-color-light-grey', 'VOL-color-light-green'
+];
+
 export default {
   props: {
     styles: {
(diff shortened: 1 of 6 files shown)


modules/volunteers/admin/sitesupport/vue.js
```diff
@@ -61,6 +66,50 @@ var app = Vue.createApp({
...
+    },
+    clearMember() {
+      this.userId = null;
+      this.userData = null;
+      this.$refs.vhtbl.load();
+      this.$refs.gfttbl.load();
+      this.$refs.lookup.clear();
+    },
+    enterAdmin() {
+      var target = '';
```
rkaiser-dev, line 78: Can we use `let` here instead of `var`?

Additionally this could be a good use case for `URLSearchParams`:

```
const searchParams = new URLSearchParams('Function=volunteers/admin');

if (this.userId) {
  searchParams.append(`volunteerId=${this.userId}`);
}

const target = `index.php?${searchParams.toString()}`;
```

modules/volunteers/admin/sitesupport/vue.js
```diff
@@ -61,6 +66,50 @@ var app = Vue.createApp({
...
+    failAdmin(error) {
+      document.getElementById('admin_slider').checked = false;
+      if (error) {
+        alertbox('Login Failed (' + error + ')');
+      }
+    },
+    toggleAdminMode() {
+      document.cookie = 'CIAB_VOLUNTEERADMIN=;expires=Thu, 01 Jan 1970 ' +
+                        '00:00:01 GMT;';
+      var target = '';
```
rkaiser-dev, line 95: Can we use `let` here instead of `var`?

Additionally this could be a good use case for `URLSearchParams`:

```
const searchParams = new URLSearchParams('Function=volunteers/admin');

if (this.userId) {
  searchParams.append(`volunteerId=${this.userId}`);
}

const target = `index.php?${searchParams.toString()}`;
```

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
     }
   },
   methods: {
+    load() {
+      this.records = null;
+      this.totalHours = 0;
+      this.totalSpentHours = 0;
+      this.hours = [];
+      this.totalVolunteers = null;
```
rkaiser-dev, line 44: Some of these values are already being set in the initial `data()` function - is there a reason why we need to set them again here?

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
...
+
+      showSpinner();
+      this.summeryColumns = [
+        {value:'name', title:'Department Worked', source:null},
```
rkaiser-dev, line 48: Do we need to set `source` if it is not using anything? Would simply omitting it be an option?

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
...
+        {value:'hours', title:'Hours', source: this.hours}
+      ];
+
+      apiRequest('GET', '/department','max_results=all')
+        .then((response) => {
+          const result = JSON.parse(response.responseText);
+          this.departments = [];
+          result.data.forEach((entry) => {
+            this.departments.push(entry);
+          });
```
rkaiser-dev, line 66: Would this be cleaner?

`this.departments = result.data;`

Does `result.data` come back with more than what we want to actually track here?

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
...
+
+          if (this.$parent.userId != null) {
+            this.columns = this.detailColumns;
+            this.records = this.hours;
+            apiRequest('GET', '/member/' + this.$parent.userId + '/volunteer/hours','max_results=all')
+              .then((response) => {
+                const result = JSON.parse(response.responseText);
+                result.data.forEach((entry) => {
+                  this.hours.push(entry);
+                })
```
rkaiser-dev, line 78: Would this be cleaner?

`this.hours = result.data;`

Does `result.data` come back with more than what we want to actually track here?

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
...
+
+      apiRequest('GET', '/department','max_results=all')
+        .then((response) => {
+          const result = JSON.parse(response.responseText);
+          this.departments = [];
+          result.data.forEach((entry) => {
+            this.departments.push(entry);
+          });
+
+          this.departments.sort((a,b) => (a.name > b.name) ? 1 : -1);
```
rkaiser-dev, line 68: Would it be worth considering adjusting the query on the backend to do this sort for us?

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
...
+                if (this.footer && this.$parent.userId != null) {
+                  apiRequest('GET', '/member/' + this.$parent.userId + '/volunteer/claims/summary','max_results=all')
+                    .then((response) => {
+                      const result = JSON.parse(response.responseText);
+                      this.totalSpentHours = result.spent_hours;
+                    })
+                    .finally(() => {
+                      hideSpinner();
+                      this.$emit('hourChange', this.totalHours, this.totalSpentHours);
+                    })
```
rkaiser-dev, line 91: Is there any reason we cannot include `spent_hours` in the API endpoint for `/member/{id}/volunteer/hours`?

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
...
+              })
+          } else {
+            this.columns = this.summeryColumns;
+            this.records = this.departments;
+            apiRequest('GET', '/event/current/volunteer/hours/summary','max_results=all')
+              .then((response) => {
+                const result = JSON.parse(response.responseText);
+                result.data.forEach((entry) => {
+                  this.hours[entry.department.id] = entry;
+                })
```
rkaiser-dev, line 105: Would this be cleaner?

`this.hours = result.data;`

It seems like we want to reset the data, is there any reason why this might not be a good approach?

modules/volunteers/sitesupport/hourTable.js
```diff
@@ -110,6 +36,90 @@ export default {
...
+                this.totalVolunteers = result.total_volunteer_count;
+                this.totalHours = result.total_hours;
+                apiRequest('GET', '/event/current/volunteer/claims/summary','max_results=all')
+                  .then((response) => {
+                    const result = JSON.parse(response.responseText);
+                    this.totalSpentHours = result.spent_hours;
+                  })
+                  .finally(() => {
+                    hideSpinner();
+                  })
```
rkaiser-dev, line 115: Is there any reason we cannot include `spent_hours` in the API endpoint for `/event/current/volunteer/hours/summary`?

rkaiser-dev (changes requested):

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -143,6 +41,106 @@ export default {
...
+            if (b.reward_group != null && a.reward_group == null) {return 1;}
+            if (b.reward_group == null && a.reward_group == null) {return 0;}
+            if (parseInt(b.reward_group.id) == parseInt(a.reward_group.id)) {
+              if (a.value == b.value) {
+                return ((a.name > b.name) ? 1 : -1);
+              }
+              return (parseFloat(a.value) > parseFloat(b.value)) ? 1 : -1;
+            }
+            return (parseInt(a.reward_group.id) > parseInt(b.reward_group.id) ? 1 : -1);
+          });
```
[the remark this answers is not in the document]
AricStewart: `===` check... 
However item.value is a float and I am seeing it being parsed as a string. 

`console.log(typeof a.value, a.value);` => `string 1.30`
Same with reward_group_id
`console.log(typeof a.reward_group.id, a.reward_group.id);` => `string 8`

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -143,6 +41,106 @@ export default {
...
+
+          if (this.$parent.userId != null) {
+            apiRequest('GET', '/member/' + this.$parent.userId + '/volunteer/claims', 'max_results=all')
+              .then((response) => {
+                const result = JSON.parse(response.responseText);
+                this.claims = result.data;
+                result.data.forEach((entry) => {
+                  var rewardId = entry.reward.id;
```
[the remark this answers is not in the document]
AricStewart: Completely removed.

modules/volunteers/admin/sitesupport/vue.js
AricStewart: This is a bit funky.

I am getting redirects looking like this `http://localhost:8080/index.php?Function=volunteers%2Fadmin&volunteerId=2001`  They appear to work but is that `%2F` going to be an issue

modules/volunteers/admin/sitesupport/vue.js
AricStewart: Same `%2f` thing as above

modules/volunteers/sitesupport/hourTable.js
AricStewart: We could, and just change the check in the code later.  Feels like six of one or half dozen of another, unless you are passionate about it.

modules/volunteers/sitesupport/hourTable.js
AricStewart: Something funky happens here...
So above you see `this.records = this.hours`  which should be assignment by reference.
if we do the whole `this.hours.push` thing, then `this.records` is correct. However if we just assign `this.hours` to `result.data` then `this.records` is not updated and remains empty.

modules/volunteers/sitesupport/hourTable.js
AricStewart: Either way, you said above that sorting is efficient in JavaScript so not sure if it is better to sort on the client or the server. I tend to like to push easy work to the client.

modules/volunteers/sitesupport/hourTable.js
AricStewart: The hours table expected the index to be the department id.  and the data coming back from the API request is not that way.

modules/volunteers/sitesupport/hourTable.js
AricStewart: Yup see above...

modules/volunteers/admin/sitesupport/giftTable.js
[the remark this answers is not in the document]
rkaiser-dev: I'm surprised by that - is the API returning numeric values as strings? If so, that makes more sense. If not, I can poke around with it a bit too, but in my experience this type of coercion hasn't been strictly necessary.

modules/volunteers/admin/sitesupport/vue.js
rkaiser-dev: The `%2F` in this case is the URL encoding, it should represent `/`. Does it still direct to the correct location? What potential issues do you foresee with the encoded value?

Another option if we want to keep the `/`' in the URL is to skip the URLSearchParams but things up a bit.

```
let target = 'index.php?Function=volunteers/admin`

if (this.userId) {
  target += `&volunteerId=${this.userId}`;
}

modules/volunteers/sitesupport/hourTable.js
rkaiser-dev: That seems weird. I can explore more into that as I dive into Vue, it could just be a quirk of the framework or something.

Another clean option might be to do something like

`this.hours.push(...result.data);` or `this.hours = [...result.data];`

modules/volunteers/sitesupport/hourTable.js
rkaiser-dev: Yeah, sorting on the client is probably fine. I have no strong preference here.

rkaiser-dev (approved):

PR #610, opened 2023-08-09
Aric/vol groups

AricStewart:
Work on better support of volunteer reward groups.

### DIFF
--- a/modules/volunteers/admin/sitesupport/editPrize.js
+++ b/modules/volunteers/admin/sitesupport/editPrize.js
@@ -1,6 +1,8 @@
 /* jshint esversion: 6 */
 /* globals  apiRequest, showSidebar, confirmbox, simpleObjectToRequest */
 
+const NEW_GROUP = -2;
+
 export default {
   data() {
     return {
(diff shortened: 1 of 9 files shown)


modules/volunteers/admin/sitesupport/editPrize.js
```diff
@@ -86,25 +85,65 @@ export default {
...
-      if (!found) {
-        this.reward_limit = -1;
       }
     },
+    editPrizeName() {
+      document.getElementById('da_box').style.display = 'inline';
+    },
+    doneEdit() {
+      document.getElementById('da_box').style.display = 'none';
+    }
```
rkaiser-dev, line 146: Are we unable to get the modal to display through the use of conditional classes/styling here? Perhaps through the use of a boolean on our model, and then apply the class conditionally using that boolean on the modal's definition?

Then we could just update the functions here to set it to true or false based on the appropriate condition (probably `true` for `editPrizeName()` and `false` for `doneEdit()`)

Does something like this not work?

```
this.isEditingPrizeName = false;

<div class="UI-Modal" v-show={{isEditingPrizeName}}>
  ...
</div>
```

rkaiser-dev (changes requested):

rkaiser-dev (approved):

PR #611, opened 2023-08-10
Upgrade eslint version

AricStewart:


rkaiser-dev (approved):
This seems fine. It runs fine with `npx eslint` and starts fine, too.

PR #612, opened 2023-08-13
Initial Changes for Splitting ConCom List into front-end/back-end

rkaiser-dev:
This PR includes:

- Fixes an issue with incorrect values for `child_count` and `email` list in Department API call.
- Adds Jest for JS testing
- Adds initial Vue component code for rendering Division Headers

I am open to moving the JS tests to be closer to where the code actually lives, but for now have included it in `<rootDir>/test/sitesupport`. I wasn't sure how or what is moved over to production servers when a production tag is cut for release.

Not included is actually rendering this - I did not want to commit those changes with this PR, but have been testing with them locally. If you'd like to see the components in action, you can add the following snippet to `<rootDir>/modules/concom/pages/body.inc` between lines 22-24, or wherever you'd like. If done there, it will render at the very bottom of the page, below the existing list of divisions and departments.

```
  <hr/>

  <div id="staff-app">
    <staff-list></staff-list>
  </div>
```

### DIFF
--- a/api/src/Controller/BaseController.php
+++ b/api/src/Controller/BaseController.php
@@ -493,7 +493,7 @@ abstract class BaseController
                 $select->subselect()
                     ->columns('COUNT(DepartmentID)')
                     ->from('`Departments` d2')
-                    ->whereEquals(['d2.ParentDepartmentID' => 'd1.DepartmentID'])
+                    ->where('d2.ParentDepartmentID = d1.DepartmentID')
                     ->andWhere("NOT NAME = 'Historical Placeholder'")
                     ->as('child_count')
                     ->getStatement()
@@ -502,13 +502,17 @@ abstract class BaseController
                 $select->subselect()
                     ->columns('GROUP_CONCAT(Email)')
                     ->from('EMails')
-                    ->whereEquals(['DepartmentID' => 'd1.DepartmentID'])
+                    ->where('DepartmentID = d1.DepartmentID')
                     ->as('email')
                     ->getStatement()
             )
             ->from('Departments d1');
         if ($id !== null) {
-            $select->where("(DepartmentID = '$id' OR Name = '$id')");
+            $select->orWhere('(')
+            ->catWhere('DepartmentID = ', $id)
+            ->catWhere(' OR ')
+            ->catWhere('Name = ', $id)
+            ->catWhere(')');
         }
         $placeholders = $select->subselect()->columns('DepartmentID')->from('Departments')->whereEquals(['Name' => 'Historical Placeholder']);
         $select->where('DepartmentID NOT IN ', $placeholders);
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,4 @@
+/* globals module */
+module.exports = {
+  presets: [ ['@babel/preset-env', { targets: { node: 'current' } }] ]
+};
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,19 @@
+/* globals module */
+const config = {
+  coverageDirectory: '/coverage',
+  coveragePathIgnorePatterns: [ '/node_modules' ],
+  coverageThreshold: {
+    global: {
+      branches: 95,
+      functions: 95,
+      lines: 95,
+      statements: 95
+    }
+  },
+  testEnvironment: 'node',
+  testMatch: [ '**/test/sitesupport/**/*.test.js' ],
+  testPathIgnorePatterns: [ '/node_modules/' ],
+  verbose: true
+};
+
+module.exports = config;
--- a/modules/concom/sitesupport/components/staff-list.js
+++ b/modules/concom/sitesupport/components/staff-list.js
@@ -1,0 +1,34 @@
+/* globals apiRequest */
+import { extractDivisionHierarchy } from '../division-parser.js';
+
+const TEMPLATE = `
+  <staff-division v-for="division in divisionHierarchy" :division=division :divisionHierarchy=divisionHierarchy></staff-division>
+`;
+
+const INIITIAL_DATA = {
+  divisionHierarchy: []
+};
+
+const fetchDivisionData = async() => {
+  const response = await apiRequest('GET', 'department');
+  const departmentData = JSON.parse(response.responseText);
+
+  return extractDivisionHierarchy(departmentData.data);
+};
+
+const onMounted = async(componentInstance) => {
+  const result = await fetchDivisionData();
+  componentInstance.divisionHierarchy.push(...result);
+};
+
+const staffListComponent = {
+  template: TEMPLATE,
+  data() {
+    return INIITIAL_DATA
+  },
+  async mounted() {
+    await onMounted(this);
+  }
+};
+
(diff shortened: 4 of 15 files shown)


jest.config.js
```diff
@@ -0,0 +1,18 @@
+const config = {
+  coverageDirectory: '/coverage',
+  coveragePathIgnorePatterns: ['/node_modules'],
+  coverageThreshold: {
+    global: {
+      branches: 95,
```
rkaiser-dev, line 6: Incidentally, after running `npx jest --coverage` locally I notice that I am not even meeting my own suggestion for this. I will take a look and see what I can do to improve branch coverage.

api/src/Controller/BaseController.php
```diff
@@ -493,7 +493,7 @@ public function getDepartment($id)
                 $select->subselect()
                     ->columns('COUNT(DepartmentID)')
                     ->from('`Departments` d2')
-                    ->whereEquals(['d2.ParentDepartmentID' => 'd1.DepartmentID'])
+                    ->where('d2.ParentDepartmentID = d1.DepartmentID')
```
AricStewart, line 496: Isn't this the same?  Does anything meaningful change here?

api/src/Controller/BaseController.php
```diff
@@ -502,13 +502,17 @@ public function getDepartment($id)
                 $select->subselect()
                     ->columns('GROUP_CONCAT(Email)')
                     ->from('EMails')
-                    ->whereEquals(['DepartmentID' => 'd1.DepartmentID'])
+                    ->where('DepartmentID = d1.DepartmentID')
```
AricStewart, line 505: Same here.

babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: [['@babel/preset-env', { targets: { node: 'current' } }]]
+};
```
AricStewart: Feels like much of this commit could be merged into previous commits as fixes.

modules/concom/sitesupport/division-parser.js
AricStewart, line 1: Again fixups?

api/src/Controller/BaseController.php
rkaiser-dev: When calling `api/department` the value of `child_count` was `null` with the original code and now correctly reflects the `child_count` with the new changes.

So unless `child_count` was supposed to be `null`, I believe there may have been an error in syntax. It could probably be achieved with `whereEquals` but I happened to just run it with `where` and got an accurate count.

api/src/Controller/BaseController.php
rkaiser-dev: When calling `api/department` the value of `email` was null with the original code and now correctly reflects the `email` list with the new changes.

Unless `email` was supposed to be null here, I believe there may have been an error in the syntax.

babel.config.js
rkaiser-dev: Taking a guess, but I'm assuming this means you'd rather the lint fixes be put onto the original commit instead?

api/src/Controller/BaseController.php
AricStewart: The statement is indeed wrong.
should be:
`->whereEquals(['DepartmentID' => d1.DepartmentID])`
There are single quotes that are bad.  Switching it to yours is not bad either but I was just curious.

api/src/Controller/BaseController.php
AricStewart: as below.  it was bad single quotes...

babel.config.js
AricStewart: That would be correct.

AricStewart: If you have ever had to do a large bisect to try to find a bug then you understand why you get twitchy that every patch should be correct and stand alone.  I would guess we are likely to never have to do that here... but I have a lot in other code bases and it really has burned into my being that need for patches to be atomically as correct as can be.

babel.config.js
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line no-undef
 module.exports = {
-  presets: [['@babel/preset-env', { targets: { node: 'current' } }]]
-};
\ No newline at end of file
+  presets: [ ['@babel/preset-env', { targets: { node: 'current' } }] ]
```
AricStewart, line 3: This line is added in **Add data parser and test** any chance this chance could be made there? instead of added one way and then changed in the next patch in the series?

jest.config.js
AricStewart: Some things added in **Add jest** that would be ideally added correctly there instead of fixed here.

jest.config.js
```diff
@@ -10,9 +10,10 @@ const config = {
...
+  testMatch: [ '**/test/sitesupport/**/*.test.js' ],
+  testPathIgnorePatterns: [ '/node_modules/' ],
   verbose: true
-}
+};
 
-module.exports = config;
\ No newline at end of file
+// eslint-disable-next-line no-undef
+module.exports = config;
```
AricStewart, line 19: Some things added in **Add jest** that would be ideally added correctly there instead of fixed here.

modules/concom/sitesupport/components/staff-list.js
```diff
@@ -0,0 +1,34 @@
+/* globals apiRequest */
+import { extractDivisionHierarchy } from '../division-parser.js';
+
+const TEMPLATE = `
+  <staff-division v-for="division in divisionHierarchy" :division=division :divisionHierarchy=divisionHierarchy></staff-division>
+`;
+
+const INIITAL_DATA = {
```
AricStewart, line 8: Spelling?

modules/concom/sitesupport/division-parser.js
```diff
@@ -65,17 +70,23 @@ export const extractDivisionHierarchy = (departmentData) => {
...
+    division.departments.sort((a, b) => a.name.localeCompare(b.name));
   });
-}
 
-// module.exports = {
-//   extractDivisionHierarchy
-// }
\ No newline at end of file
+  return sortedDivisions;
+}
```
AricStewart, line 92: Again revisions to **Add data parser and test** could it get split out and integrated there?

AricStewart:
I will admit that splitting and merging patches is an art form that take practice.  If you have hit frustration limit then I can let it slide but since most of the splitting affect whole new modules it should not be hard.

To remove changes from a given commit

`git show <sourcefile> > <outfile>`
`patch -p1 -R < <outfile>` 
`git add <sourcefile>`
`git commit --amend`

then

`patch -p1 < <outfile>`
`git add <sourcefile>`
`git commit`

will add those changes by themselves as a separate patch that can be moved around and merged using

`git rebase -i origin/master`

Then you can pull the changes out of you last patch and merge them into the other and things would look great!

rkaiser-dev:
Sure, I can make the updates to split out the changes a bit better. No problem. I'll try to knock that out today or tomorrow.

rkaiser-dev:
I think I've managed to resolve all of the feedback and cleaned up the git history appropriately.

AricStewart (approved):

PR #613, opened 2023-08-14
Allow sorting of volunteer prize table

AricStewart:


modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -260,6 +244,57 @@ export default {
...
+        this.decendingSort = !this.decendingSort;
+      }
+      this.sortColumn = column.value;
+      let targetLeft = (this.decendingSort) ? 1 : -1;
+      let targetRight  = (this.decendingSort) ? -1 : 1;
+
+      this.records.sort((a,b) => {
+        let aValue = a[column.value];
+        let bValue = b[column.value];
+        if (column.value == 'remaining') {
```
rkaiser-dev, line 262:
```suggestion
        if (column.value === 'remaining') {
```

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -260,6 +244,57 @@ export default {
...
+    sortTable(column) {
+      if (column.unsortable) {
+        return;
+      }
+      if (column.value === this.sortColumn) {
+        this.decendingSort = !this.decendingSort;
+      }
+      this.sortColumn = column.value;
+      let targetLeft = (this.decendingSort) ? 1 : -1;
+      let targetRight  = (this.decendingSort) ? -1 : 1;
```
rkaiser-dev, line 257: I think we might be able to simplify this a little bit for ascending/descending sort. Let me know what you think.

First, it seems like we're trying to determine whether we're sorting ascending or descending order.

We might be able to use something like `const sortDirection = this.descendingSort ? -1 : 1;`

Then, we can take the values of our return statements and multiply them by the sort direction to get ascending or descending order.

Line 272 could become: `return (parseFloat(a.value) - parseFloat(b.value)) * sortDirection;`
Line 274 could become `return a.name.localeCompare(b.name) * sortDirection;`
Line 276 could become `return (aValue - bValue) * sortDirection;`

So as an example, let's imagine we're doing ascending order sort - I will just cover line 272.

a.value = 10
b.value = 5

(10 - 5) * 1 = 5, so `a` should be sorted lower than `b`.

c.value = 5
d.value = 7

(5 - 7) * 1 = -2, so `c` should be sorted higher than `d`.

If we flip the `sortDirection` value to `-1` we should get the opposite ordering.

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -260,6 +244,57 @@ export default {
...
+          bValue = parseFloat(bValue);
+        }
+
+        if (a.reward_group !== null && b.reward_group === null) {return -1;}
+        if (b.reward_group !== null && a.reward_group === null) {return 1;}
+        if (b.reward_group === null && a.reward_group === null) {return 0;}
+        if (parseInt(b.reward_group.id) === parseInt(a.reward_group.id)) {
+          return ((aValue > bValue) ? targetLeft : targetRight);
+        }
+        return (parseInt(a.reward_group.id) > parseInt(b.reward_group.id) ? 1 : -1);
```
rkaiser-dev, line 296: It looks like the main different here on this sort is the presence of the reward groups for each entry, correct? Are these something that we would also want to sort in ascending/descending order? Or is it only the items within the groups that sort this way?

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -260,6 +244,57 @@ export default {
...
+          }
+          return ((a.name > b.name) ? targetLeft : targetRight);
+        }
+        return ((aValue > bValue) ? targetLeft : targetRight);
+      });
+
+      this.records.sort((a,b) => {
+        let aValue = a[column.value];
+        let bValue = b[column.value];
+        if (column.value == 'remaining') {
```
rkaiser-dev, line 282:
```suggestion
        if (column.value === 'remaining') {
```

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -260,6 +244,57 @@ export default {
...
+        } else if (column.type === Number) {
+          aValue = parseFloat(aValue);
+          bValue = parseFloat(bValue);
+        }
+
+        if (a.reward_group !== null && b.reward_group === null) {return -1;}
+        if (b.reward_group !== null && a.reward_group === null) {return 1;}
+        if (b.reward_group === null && a.reward_group === null) {return 0;}
+        if (parseInt(b.reward_group.id) === parseInt(a.reward_group.id)) {
+          return ((aValue > bValue) ? targetLeft : targetRight);
```
rkaiser-dev, line 294: If this is the only piece that should be sortable, consider the comment above about `sortDirection` and we could simplify this to `return (aValue - bValue) * sortDirection;`

rkaiser-dev (changes requested):

AricStewart:
Great suggestions! I was able to really clean up the sorting function. Thanks!

modules/volunteers/admin/sitesupport/giftTable.js
```diff
@@ -260,6 +244,43 @@ export default {
...
+
+      this.records.sort((a,b) => {
+        let result = (parseFloat(a[column.value]) - parseFloat(b[column.value])) * sortDirection;
+        if (column.value === 'remaining') {
+          result = ((a.inventory - a.claimed) - (b.inventory - b.claimed)) * sortDirection;
+        } else if (column.type === String) {
+          result = a[column.value].localeCompare(b[column.value]) * sortDirection;
+        }
+
+        if (result === 0.0) {
```
rkaiser-dev, line 268: Curious whether this may give us trouble down the road due to floating point precision. I guess if it comes up, maybe we address it at that time. I think some eventual options might be things like `Math.floor()` or `Number.toPrecision()`

rkaiser-dev (approved):

PR #619, opened 2023-09-13
api: refine the staff listing methods

AricStewart:
* Move /department/staff to /staff as it lists all staff regardless of department.
* Rename `GetDepartment` to `ListDepartmentStaff` as it is much more descriptive and correct
* add `subdepartments` parameter to `ListDepartmentStaff` to list staff in sub-departments of that department (division) is requested

### DIFF
--- a/api/src/Modules/staff/App/Routes.php
+++ b/api/src/Modules/staff/App/Routes.php
@@ -18,6 +18,7 @@ function setupStaffAPI($app, $authMiddleware)
     $app->group(
         '/staff',
         function () use ($app, $authMiddleware) {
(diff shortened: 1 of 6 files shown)


rkaiser-dev:
This looks good - I verified with some limited testing locally while the server was spun up and it gave me the results I expected. When passing a department ID that represented a division and setting the query param, I was getting all division staff as expected.

When passing a department ID that represented a department it was getting the department staff as expected. I also tried to play with the query param here and continued to get the expected results.

Thanks!

rkaiser-dev (approved):

PR #620, opened 2023-09-15
api: Implement `/member/{id}/permissions`

AricStewart:


rkaiser-dev (commented):
This seems to get some permissions back from the API endpoint, but it seems like the ConCom ones aren't being added.

I do see that volunteer roles are coming through, so I wonder if it's maybe just something with ConCom permissions specifically not being triggered to get added to this system.

It could also be that my database doesn't have any ConCom permissions either, though I'm not quite sure the best way to verify that.

rkaiser-dev:
Thanks for the detailed explanation. This helps. I think this should be fine as-is and we can play around with making sure all of the other permissions are also correctly added later. I'll approve this for now, and will plan to poke at it some more once I'm ready to make those optimizations :) I appreciate your help with this!

rkaiser-dev (approved):

PR #621, opened 2023-09-26
Ryan/concom sidebar

rkaiser-dev:
PR Includes:
- Provides "Edit" button in UI for users with appropriate permissions
- Populates sidebar with staff member details when "Edit" clicked
- Allows updating staff member note and position
- Allows removal of staff member from department

Made as much effort as possible to keep API requests low. Editing or removing staff members modifies the list of staff members that we have collected on page load rather than making additional requests to perform updates.

Please note that the ordering of these users does not change when modifying in place like this. I did attempt to do some sorting on the list when editing, but was not successful - additional effort may lead to results but for now I accepted this behavior once implemented.

Because the sidebar is rendered by the `staff-list.js` component, additional properties were added that have been injected for use by other components. Also because this was dragging on a bit longer than I intended, and I'd like to focus on some other areas that will help us, I chose to `watchEffect()` here for the division-level staff arrays instead of opting to use the built-in `Suspense` component that we could leverage here due to use of asynchronous calls.

I tried to keep the PR on the small side but I am seeing that it got a bit bigger than intended. I could potentially break out the behavior for API updates and the UI updates into separate PRs.

As with other PRs so far, if testing in your own UI, you will need to remove the entire contents of modules/concom/pages/body.inc and then add:

<div id="staff-app">
  <staff-list></staff-list>
</div>

In addition, you will need to uncomment the imports and components in modules/concom/sitesupport/concom_v2.js

### DIFF
--- a/api/src/Modules/staff/Controller/PutStaffMembership.php
+++ b/api/src/Modules/staff/Controller/PutStaffMembership.php
@@ -1,0 +1,102 @@
+<?php declare(strict_types=1);
+/*.
+    require_module 'standard';
+.*/
+
+/**
+ *  @OA\Put(
+ *      tags={"members"},
+ *      path="/member/{id}/staff_membership",
+ *      summary="Update a staff membership for the member",
+ *      @OA\Parameter(
+ *          description="The ID of the member",
+ *          in="path",
+ *          name="id",
+ *          required=true,
+ *          @OA\Schema(
+ *              description="Member ID",
+ *              type="integer"
+ *          )
+ *      ),
+ *      @OA\RequestBody(
+ *          @OA\MediaType(
+ *              mediaType="application/x-www-form-urlencoded",
+ *              @OA\Schema(
+ *                  @OA\Property(
+ *                      property="Department",
+ *                      description="ID or name of the department",
+ *                      nullable=false,
+ *                      type="string"
+ *                  ),
+ *                  @OA\Property(
+ *                      property="Position",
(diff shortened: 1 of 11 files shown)


api/src/Modules/staff/Controller/PutStaffMembership.php
```diff
@@ -0,0 +1,107 @@
...
+            throw new InvalidParameterException('Body required');
+        }
+
+        if (!array_key_exists('Department', $body)) {
+            throw new InvalidParameterException('Required \'Department\' parameter not present');
+        }
+
+        if (!array_key_exists('Position', $body)) {
+            throw new InvalidParameterException('Required \'Position\' parameter not present');
+        }
```
AricStewart, line 83: There is a new function that can be called instead of a bunch of array_key_exists

`$request->checkRequiredBody`

See if that works for you. I have used it in the newer implementations but have not gone back and replaced it other places.

AricStewart:
ok my Javascript skills are still maturing but what I am seeing on the javascript side looks great!  So I guess just check in on that one part on the API side and things look great!

api/src/Modules/staff/Controller/PutStaffMembership.php
rkaiser-dev: I made the change here and it works great, thanks for the tip! Let me know if the changes look better to you.

AricStewart (approved):

PR #622, opened 2023-10-17
First pass restructure RBAC

AricStewart:
Require all the pieces of the API that check RBAC permissions to register those permissions in a unified way. This will give us some more flexibility in future shifts in RBAC frameworks.

What do you think of a starting framework like this?

### DIFF
--- a/api/src/Controller/System/BaseSystem.php
+++ b/api/src/Controller/System/BaseSystem.php
@@ -26,5 +26,18 @@ abstract class BaseSystem extends BaseController
     }
 
 
+    public static function install($database): void
+    {
+
+    }
+
+
+    public static function permissions($database): ?array
+    {
+        return ['api.get.log', 'api.get.configuration', 'api.put.log', 'api.put.configuration'];
+
+    }
+
+
     /* End BaseSystem */
 }
--- a/api/src/Controller/System/BaseSystem.php
+++ b/api/src/Controller/System/BaseSystem.php
@@ -26,7 +26,7 @@ abstract class BaseSystem extends BaseController
     }
 
 
-    public static function install($database): void
+    public static function install($container): void
     {
 
     }
--- a/api/src/Modules/staff/Controller/BaseStaff.php
+++ b/api/src/Modules/staff/Controller/BaseStaff.php
@@ -155,6 +155,36 @@ abstract class BaseStaff extends BaseController
     }
 
 
+    public static function install($database): void
+    {
+
+    }
+
+
+    public static function permissions($database): ?array
+    {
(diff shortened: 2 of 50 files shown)


api/src/Controller/System/BaseSystem.php
```diff
@@ -26,5 +26,18 @@ public function __construct(Container $container)
...
 
+    public static function install($container): void
+    {
+
+    }
+
+
+    public static function permissions($database): ?array
+    {
+        return ['api.get.log', 'api.get.configuration', 'api.get.log', 'api.put.configuration'];
```
rkaiser-dev, line 37:
```suggestion
        return ['api.get.log', 'api.get.configuration', 'api.put.log', 'api.put.configuration'];
```

Looks like you might have meant `api.put.log` on the second one of these in the array.

api/src/Modules/staff/Controller/BaseStaff.php
```diff
@@ -155,6 +155,36 @@ public function __construct(Container $container)
...
+        $result = ['api.delete.staff.all', 'api.get.staff', 'api.post.staff.all', 'api.put.staff.all',
+            'api.delete.staff.all'];
+
+        $values = Select::new($database)
+            ->columns('DepartmentID')
+            ->from('Departments')
+            ->orderBy('`DepartmentID` ASC')
+            ->fetchAll();
+
+        foreach ($values as $value) {
```
rkaiser-dev, line 175: Will this allow for the hierarchy where Division members can make changes to any department member within that division?

I'll admit that when I was working to add the RBAC to `ModuleStaff.php`, I more or less copied the logic that was happening in the ConCom RBAC to this for Divisions/Departments to make sure everything was the same as before.

rkaiser-dev (commented):
This seems like it is probably trending in the right direction. I think it might be good to eventually supplement with a 3rd-party module for RBAC, I seem to recall PbpRbac mentioned for this and based on a quick glance I think that would be fine too.

Ideally, the roles and permissions get stored in a database or some kind of file so we don't have to build them every time, although I understand that this will be a multiple step process either way.

AricStewart:
I should have you look at this once more before I keep moving forward.

The next steps are to pull the RBAC initilzation that is happening in `modules/concom` and put them in `api/src/Modules/staff` and have the modules/concom side refer to the API side instead of vice versa for now.

Also then to go through all the places in `modules` where we are using the rbac directly and have them go through APIs and remove from the PHP side of the client all together.

After that.  Look to replace Vend/Lammas with another backend.

rkaiser-dev (commented):
Overall looks like it's still good progress. I didn't notice anything out of place in the new changes.

PR #623, opened 2023-10-18
Correction to Member RBAC checks

AricStewart:


modules/concom/functions/RBAC.inc
```diff
@@ -199,11 +199,11 @@ class ConComRBAC
...
         $positions = POSITION::getConComPosition($accountId);
         foreach ($positions as $pos) {
-            $value = $pos['departmentId'].'.'.$pos['positionId'];
+            $role = $pos['departmentId'].'.'.$pos['positionId'];
             try {
                 $result = array_unique(array_merge($result, self::instance()->rbac->getRole(
                     strval($role)
-                )->getPermissions($children)));
+                )->getPermissions(true)));
             } catch (Exception\InvalidArgumentException $e) {
```
rkaiser-dev, line 207: Ah yeah, I did run into this issue when I was trying it out locally with edits, but hadn't committed any fix yet.

It looks like this code snippet: `self::instance()->rbac->getRole(strval($role))->getPermissions(true)));` might be the same logic that is happening in `self::getPermissions()`. The only difference is that function seems to return a `null` if an error happens, where in our case we are just ignoring the error.

When I solved for this locally, I ended up doing something like this:

```
try {
  $result = array_unique(array_merge($result, self::getPermissions($role)));
} catch (Exception\InvalidArgumentException $e) {
}
```

I think what you have here will work just fine, too :)

rkaiser-dev (approved):
Looks good to me, thanks for adding some tests around this as well!

PR #625, opened 2023-10-21
Additional ConCom List Updates

rkaiser-dev:
This PR includes:

- Adds RBAC to ModuleStaff in the API.
- Updates to match parity on user permissions for adding members and editing members.

What should be true for each of the positions, based on permissions:

- Admin User
  - Can add to any division or department
  - Can edit anyone in any division or department
- Division Director
  - Can add to the current division or any department within that division
  - Can edit Division Support in current division, and anyone in any department within that division
- Division Support
  - Can add to any department within that division
  - Can edit anyone in any department within that division
- Department Head
  - Can add to their department
  - Can edit sub-heads and specialists within their department
- Department Sub-Head
  - Cannot add to their department
  - Cannot edit anyone within their department
- Department Specialist
  - Cannot add to their department
  - Cannot edit anyone within their department

Next up, for future PRs:
- Need to confirm the behavior of the Sidebar in old vs new. Want to see if the RBAC stuff limits what shows in the dropdown (not sure if it does or doesn't). For example, can or should a department head be able to add a new department head in their department? Based on edit logic, I'd guess no.
- Fix rendered format for special divisions.
- Separate Organization Donut data from UI

### DIFF
--- a/api/src/Modules/staff/ModuleStaff.php
+++ b/api/src/Modules/staff/ModuleStaff.php
@@ -8,6 +8,7 @@ namespace App\Modules\staff;
 use App\Modules\BaseModule;
 use Slim\Http\Request;
 use Slim\Http\Response;
+use Atlas\Query\Select;
 
 class ModuleStaff extends BaseModule
 {
@@ -16,6 +17,13 @@ class ModuleStaff extends BaseModule
     public function __construct($source)
     {
         parent::__construct($source);
+
+    }
+    
+
+    public function install($container)
+    {
+        $container->RBAC::customizeRBAC('\App\Modules\staff\ModuleStaff::customzieStaffRBAC');
 
     }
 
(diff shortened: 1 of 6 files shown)


api/src/Modules/staff/ModuleStaff.php
```diff
@@ -40,5 +42,76 @@ public function handle(Request $request, Response $response, $data, $code)
     }
 
 
+    public function customizeAnnouncementRBAC($instance, $database)
+    {
```
AricStewart, line 46: As above likely `customizeStaffRBAC`

api/src/Modules/staff/ModuleStaff.php
```diff
@@ -40,5 +42,76 @@ public function handle(Request $request, Response $response, $data, $code)
...
+        $role = $instance->getRole('all.staff');
+        $role->addPermission('api.get.staff');
+
+        if (array_key_exists('IS_ADMIN', $_SESSION) && $_SESSION['IS_ADMIN']) {
```
AricStewart, line 50: I am not sure if we should be checking IS_ADMIN here... IS_ADMIN is going to basically bypass all this logic anyway.  so we should not add anything to an admin role. 

Also you do not use `getRole` anymore

api/src/Modules/staff/ModuleStaff.php
```diff
@@ -40,5 +42,76 @@ public function handle(Request $request, Response $response, $data, $code)
...
+            $adminRole = $instance->getRole('admin');
+            $adminRole->addPermission('api.post.staff.all');
+            $adminRole->addPermission('api.put.staff.all');
+            $adminRole->addPermission('apu.delete.staff.all');
```
AricStewart, line 54: All of this is the 'old' way...  What you want is to be doing 

`$instance->grantPermission($role, $perm);`

api/src/Modules/staff/ModuleStaff.php
```diff
@@ -40,5 +42,76 @@ public function handle(Request $request, Response $response, $data, $code)
...
+            $parentDept = $value['ParentDepartmentID'];
+            $dept = $value['DepartmentID'];
+
+            if ($parentDept != $dept) {
+                $targetRole = "$parentDept.".end(array_keys($positions));
+                try {
+                    $role = $instance->getRole($targetRole);
+                    $role->addPermission("api.put.staff.$dept.1");
+                    $role->addPermission("api.delete.staff.$dept.1");
+                } catch (Exception\InvalidArgumentException $e) {
```
AricStewart, line 84: Like above using `grantPermission` will not keep commenting here.

AricStewart (changes requested):

api/src/Modules/staff/ModuleStaff.php
rkaiser-dev: I'm not so sure about that - without this addition in `ModuleStaff`, the admin user was not getting those permissions back when calling the API for member permissions, so their super-user abilities on this page were not rendered properly even if the API would have allowed them to make the request.

api/src/Modules/staff/ModuleStaff.php
```diff
@@ -16,6 +17,7 @@ class ModuleStaff extends BaseModule
     public function __construct($source)
     {
         parent::__construct($source);
+        \ciab\RBAC::customizeRBAC('\App\Modules\staff\ModuleStaff::customizeAnnouncementRBAC');
```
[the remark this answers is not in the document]
rkaiser-dev: Yup, makes sense. All of these changes were implemented locally prior to the RBAC changes being merged. I will double check this against the current branch on main, but wanted to at least get this stuff up before then after verifying the behavior with the changes as-is.

I will rebase against the main branch and make any necessary adjustments. I will also make sure that I can actually login and don't get the redirect loop.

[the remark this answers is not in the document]
AricStewart: Sounds good. Note that the parameters to the customize function has changed so right now your code will not work.

api/src/Modules/staff/ModuleStaff.php
AricStewart: With the current code if you register the permission then the admin use will have them in their list.

modules/concom/functions/RBAC.inc
```diff
@@ -272,7 +272,7 @@ class ConComRBAC
             if (!$r) {
                 self::addRole($role, $parents);
             } else {
-                foreach($parents as $parent) {
+                foreach ($parents as $parent) {
```
rkaiser-dev, line 275: This was something my linter caught after rebasing on main.

rkaiser-dev:
I think everything should be resolved now and brought up to parity, including the new changes to RBAC. I made a couple of adjustments to the BaseStaff to make sure the permissions line up because there were some items that wouldn't be used (`api.get.staff.{deptId}`) or needed to be updated (`api.put/delete.staff.{deptId}` -> `api.put/delete.staff.{deptId}.{positionId}`) because of position-based permission requirements.

AricStewart (approved):

PR #629, opened 2023-10-24
Use API RBAC in javascript

AricStewart:
This is mostly stop-gap until things are more properly VUE-ized. It is likely we may still use some of this in the `pre.inc` pages to block sensitive pages from even loading.

Also nothing was touched in the `concom` module until rewrite and cleanup is finished.

### DIFF
--- a/functions/functions.inc
+++ b/functions/functions.inc
@@ -16,6 +16,10 @@ if (is_file($BASEDIR.'/.env')) {
 require_once(__DIR__."/../backends/email.inc");
 require_once(__DIR__."/database.inc");
 
+use \GuzzleHttp\Client;
+use \GuzzleHttp\Psr7\Request;
+use \GuzzleHttp\Exception\RequestException;
+
 // Main setup function
 $MODULENAME  = "Member Portal";
 
@@ -440,5 +444,49 @@ function core_pages()
     $valid = array_unique(array_merge($valid, $result), SORT_STRING);
 
     return $valid;
+
+}
+
+
+function RBACApiRequest($rbac)
+{
+    global $_SESSION;
+
+    $target = '/permissions/meta/'.$rbac;
+
+    if (array_key_exists('API_SERVER', $_ENV)) {
+        $url = $_ENV['API_SERVER'].'/api/'.$target;
+    } elseif (isset($WEBSERVER)) {
+        $url = $WEBSERVER.'/api/'.$target;
+    } else {
+        $url = 'http://localhost//api/'.$target;
+        ;
+    }
+    $client = new Client();
+    try {
+        $request = new Request('GET', $url, ['Authorization' => 'Bearer '.$_SESSION['apiAuth']]);
+        $response = $client->send($request);
+    } catch (RequestException $e) {
+        $response = $e->getResponse();
+        error_log('API request failed: '.$response->getBody());
+        $target = 'index.php?Function=refresh&return='.urlencode($_SERVER['REQUEST_URI']);
+        header("Location: $target");
+        return null;
+    }
+    $data = json_decode($response->getBody());
+    return $data->allowed;
+
+}
+
+
+function PageAllowedByRBAC($rbac)
+{
+    $allowed = RBACApiRequest($rbac);
+    if ($allowed === null) {
+        $target = 'index.php?Function=refresh&return='.urlencode($_SERVER['REQUEST_URI']);
+        header("Location: $target");
+    } elseif (!$allowed) {
+        goSite();
+    }
 
 }
--- a/api/src/App/Routes.php
+++ b/api/src/App/Routes.php
@@ -63,6 +63,7 @@ function setupAPIRoutes(App $app, $authMiddleware)
             $app->get('/method/announcement[/{method}[/{department}]]', 'App\Controller\Permissions\AnnouncementMethod');
 
             $app->get('/generic/{resource}/{method}[/{detail}]', 'App\Controller\Permissions\GenericResource');
+            $app->get('/meta/{permission}', 'App\Controller\Permissions\MetaResource');
         }
     )->add(new \App\Middleware\CiabMiddleware($app))->add($authMiddleware);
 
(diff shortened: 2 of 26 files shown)


functions/functions.inc
```diff
@@ -442,3 +447,45 @@ function core_pages()
...
+
+function PageAllowedByRBAC($rbac)
+{
+    $allowed = RBACApiRequest($rbac);
+    if ($allowed === null) {
+        LogoutUser();
+        goSite();
+    } elseif (!$allowed) {
+        goSite();
+    }
```
rkaiser-dev, line 489: Can you help me understand some of the decisions that have been made here? There is probably some context I don't have regarding the OAuth implementation.

- From what I gathered while poking around on my localhost, I noticed that we don't seem to make use of the refresh token portion of OAuth. Is there a specific reason we decided against this?
- I also noticed that, at least locally, the session tokens are set to expire after an hour. Is this just the default, or did we explicitly choose a 1 hour session time?

Now for the actual logic here:

- Will the user be notified when they are force-logged-out that their session had expired?
- Could this cause loss of work if someone's token expires before they're able to enter data?

functions/functions.inc
```diff
@@ -442,3 +447,45 @@ function core_pages()
     return $valid;
 
 }
+
+
+function RBACApiRequest($rbac)
```
rkaiser-dev, line 452: Can you help me understand the motive behind moving this to the PHP server calling the API it is exposing itself? 

It seems like this is doing authorization with extra steps. Could we not just check that the session user has permission to a given area of the site and deny them based on that? Or is this specifically to handle the case that a user has an OAuth token that may have expired?

Or are we planning to move out authentication entirely so that it can function as its own service?

AricStewart: The idea is to fully move all the RBAC code to the API side of the world.  From past experience, on other projects,  I always try to have layer separation enforced as much as I can, Just to help with possible API server being seperate from a possible load balanced front end, and yada yada. So I try to avoid calling directly into the API layer from the front end and instead do internal API calls.

functions/functions.inc
AricStewart: Ok this is now updated to handle renewals.... it is not ideal and maybe you know a better way to get that renewal to happen and then get the JavaScript data down into PHP.

rkaiser-dev (commented):
Let's hold off on the token refresh logic for now. It seems like the implementation could get messy especially if the PHP Backend is the thing making calls to the PHP API. Ideally, the PHP Backend can just use the RBAC logic directly until we have transitioned more of this to the UI.

There may be other approaches we can leverage in the future once we get everything to that point.

I think the changes to introduce the RBAC implementation in the API are OK, but it might be worth holding on moving the non-API split code from doing an API request approach for RBAC permissions at the moment.

AricStewart:
Sorry no  responding directly inline but it is all mixed together in my brain.

First off I am very uncomfortable with our 2 RBAC implementation, one API and one frontend PHP.  This is just a mismatch waiting to happen and really feels like it is going to be a source of pain and confusion.  I want 1 RBAC implementation and I feel like the correct place for that is in the API.

Second, trying to call the API implementation of the RBAC directly in PHP is... messy at best and likely really hard. I am trying to figure out if we can do it without having to spool up a full slim framework container. 

Third, we already have the case where the php level oauth token stored in `$_SESSION['apiAuth']` expires before the javascript one, resulting in strangely missmatched state where some parts of the system will load but others do not. 

I believe we can fix the general slow-down by caching the RBAC checks on the PHP side.  That feels like a good way to try to fix that for now.   Yes the number of back in round trips is troubling but I am not sure of another good way to do it.

rkaiser-dev:
> First off I am very uncomfortable with our 2 RBAC implementation, one API and one frontend PHP. This is just a mismatch waiting to happen and really feels like it is going to be a source of pain and confusion. I want 1 RBAC implementation and I feel like the correct place for that is in the API.

I agree, I think we're on the same page here. To make sure I understand, "frontend PHP" means, the parts of the UI that _are not_ using the API currently?

> Second, trying to call the API implementation of the RBAC directly in PHP is... messy at best and likely really hard. I am trying to figure out if we can do it without having to spool up a full slim framework container.

I agree here, too. This is why I was suggesting that we let the API stuff do it's thing and serve the front-end and then the rest of the older legacy stuff continues to do its thing, whether through the new implementation (directly, not via API calls) or otherwise. Especially because it has to spin up all of the roles again and everything else, since we aren't storing them anywhere, I think it drastically impacts performance overall. It's not a _permanent_ recommendation, but one that is just temporary until we get more of this migrated over to JS and the APIs.

> I believe we can fix the general slow-down by caching the RBAC checks on the PHP side. That feels like a good way to try to fix that for now. Yes the number of back in round trips is troubling but I am not sure of another good way to do it.

When you say caching the RBAC checks, do you mean "I've seen this user and I know they have role XYZ, so they're good"? For reasons, I would recommend against this. We would then be introducing difficult to debug behavior for our site and potentially allowing or restricting access to site functions in an inaccurate way.

If you mean "generating the roles based on positions and building up the RBAC model", then yes I do think we should cache that if we aren't already, or find a better long-term solution such as the use of a database for this. This might help with performance overall since we currently don't have them stored anywhere else, and are re-generating all of the roles every single time we are answering requests. I don't think we'd want to cache the actual checks to see whether a user can or cannot access things. 

The other option here, which would mean fewer round trips, would be to let the two implementations live separately for now and prioritize moving all of the remaining sections of the site that are not currently mostly leveraging JS and Vue to do exactly that. It's definitely risky, and there will be bugs or misses if we are doing other things with RBAC in the meantime.

It does seem like the best end-result is:
- RBAC implementation lives in the API
- JS calls into the API to do its thing
- The end

Right now, with stuff in an in-between state with some Vue-based components using the API and PHP-based data/HTML rendering, I think our RBAC implementation is similarly stuck. If we try to force the non-Vue parts of the site to use the API-based implementation, I think that will definitely be a source of pain and confusion like you've said. 

Would it be a better use of our time to get this to a state where it is acceptable (perhaps that state already exists w/o these changes) and serves the API and non-API portions of the site correctly, and then focus on getting the remaining sections onto Vue-based components that use the API before coming back to remove this stuff?

PR #634, opened 2023-11-09
RFC: Shifting DB initialization to API

AricStewart:
I believe we should shift the DB initialization from the front end to the API side.
What do you think of this approach? 
It also provides a framework to allow for a UI to re-initialize a module or of modules add records to existing tables for those to be initialized with the module potentially at a later time.  This is improtatn for things like the artshow module I am working on.

Thoughts?


rkaiser-dev:
I'm not sure I fully understand what is being done here. It seems like the DB code is being moved to the API folders rather than its current position. And then it seems like the goal is so that we have the ability to enable or disable features dynamically, rather than basing it on whether certain code is deployed or not. Is that accurate?

> I believe we should shift the DB initialization from the front end to the API side.

I also want to make sure I understand the terminology being used here. When I hear "front end" I think of UI code, so I'm guessing it means something different in this context unless I've misunderstood how things were initialized before this change.

AricStewart:
> I'm not sure I fully understand what is being done here. It seems like the DB code is being moved to the API folders rather than its current position. And then it seems like the goal is so that we have the ability to enable or disable features dynamically, rather than basing it on whether certain code is deployed or not. Is that accurate?
> 
> > I believe we should shift the DB initialization from the front end to the API side.

Yes, we do have a `DISABLEDMODULED` configuration that sort of does this but in a very rough way.

> 
> I also want to make sure I understand the terminology being used here. When I hear "front end" I think of UI code, so I'm guessing it means something different in this context unless I've misunderstood how things were initialized before this change.

Ok guilty here.  So I think of the Javascript, Vue and part of the PHP code that is all running to display the pages as the front end.  The idea is to continue to work to remove more and more of the PHP parts of that and have it more and more client side.   Front end being the part facing the user.

The Back end would be the RESTful API and parts of the code that is used to process the requests that the front ends makes  on behalf of the user. 

The PHP mess that the code original was had no such distinction. so there is server side rendered PHP all over that does things like database initialization mixed right in with all the stuff that generated the actual HTML.

A lot of this is also driven by complicated modules like the Art Show one I am working on which have initialization into existing tables and not just into their own unique tables.  That initialization cannot happen with the current framework as tables are only initialized on creation, but the module that created them first.

Instead of working on fixing that in place it seemed like a great opportunity to do a bit of general cleanup on the location and "ownership" of that part of the process.

rkaiser-dev:
> So I think of the Javascript, Vue and part of the PHP code that is all running to display the pages as the front end. 

OK this makes sense to me then. I was not considering those PHP files to be part of the front end but in this context I understand it now.

I do think that continuing to separate out the parts that are actually rendered and viewed by the user (eventually just the JS and Vue based HTML) from the backend behavior (actually things handled by the server such as fetching data from the database) as much as possible makes sense.

I am kind of curious about the strategy to have modules _add to_ existing tables vs having their own, but I think that's something that was probably decided with good reason at the time and doesn't block my ability to approve this as-is.

tools/test_prereqs.php
```diff
@@ -14,22 +17,70 @@
...
-$delete->from('AnnualCycles')->perform();
-$delete->from('AccountConfiguration')->perform();
-$delete->from('Members')->perform();
-$delete->from('HourRedemptions')->perform();
-$delete->from('VolunteerHours')->perform();
-$delete->from('VolunteerRewards')->perform();
+try {
+    $delete->from('Announcements')->perform();
+} catch (Exception $e) {
+}
```
rkaiser-dev, line 23: I know this is just a useful tool for us, but I wonder if we could make a function and call it instead of using lots of try/catch blocks like this. I'm not sure if we can build functions outside of classes in PHP, the way we can with something like JS.

It seems like this is just cleaning up data within tables, is that correct?

```
private function __cleanupTable($table): void
{
  try {
      $delete->from($table)->perform();
  } catch (Exception $e) {
  }
}
```

And then we could even put our table names in an array and just `foreach` on it to delete them all with fewer lines of code.

Let me know if this makes sense to do, and if it doesn't that's also OK to leave this as-is.

rkaiser-dev (approved):
I do have one other question. Is there any reason that we _shouldn't_ just add all of the tables and fields needed by all of our modules, given that we should only really be accessing those tables and fields if the module is enabled?

That might also help to keep things cleaner, but I'm not sure if there would be any adverse effects on other parts of the application in that approach. It _seems_ like it should be OK that way, but if I've missed something feel free to call it out.

rkaiser-dev (approved):

PR #635, opened 2023-11-10
ConCom List Cleanup

rkaiser-dev:
This PR is primarily focused on deleting code that appears unused in the PHP files now that we have switched the ConCom List UI to use Vue and the APIs.

I verified that the behavior of the ConCom List page is still functional after removing this code. I also spent some time playing around with the ConCom Structure page and confirmed that these changes did not impact anything on that page.

I do want to mention some odd behavior I noticed on the structure page, which is that we cannot add new Divisions and there are issues adding a new email address to a new Division which may be related to each other. There's an error somewhere in the SQL code to add it. 

I did not spend a ton of time investigating, other than to see if the behavior was the same on `main` as my branch, and it was, so I believe the behavior is not related to these changes. Departments can still be added, removed, and moved between Divisions.

I also have one question:
- Do we currently have any departments in production using a "Fallback Parent"? I believe that I did not bring that behavior to the ConCom List dorgnut if that is the case, but at least locally it didn't seem like anything was impacted there either.


AricStewart:
> This PR is primarily focused on deleting code that appears unused in the PHP files now that we have switched the ConCom List UI to use Vue and the APIs.
> 
> I verified that the behavior of the ConCom List page is still functional after removing this code. I also spent some time playing around with the ConCom Structure page and confirmed that these changes did not impact anything on that page.
> 
> I do want to mention some odd behavior I noticed on the structure page, which is that we cannot add new Divisions and there are issues adding a new email address to a new Division which may be related to each other. There's an error somewhere in the SQL code to add it.

I am seeing the error on adding Divisions with the current `main`, it is also present in `production` So the breaking change is likely from earlier.

> 
> I did not spend a ton of time investigating, other than to see if the behavior was the same on `main` as my branch, and it was, so I believe the behavior is not related to these changes. Departments can still be added, removed, and moved between Divisions.
> 
> I also have one question:
> 
> * Do we currently have any departments in production using a "Fallback Parent"? I believe that I did not bring that behavior to the ConCom List dorgnut if that is the case, but at least locally it didn't seem like anything was impacted there either.


Fallback parent is only used for Divisions, not departments. 

Screenshot 2023-11-10 at 9 52 07 AM

You can notice the arrows on the dogrnut, that is generated by the fallback parent for the division.

rkaiser-dev:
> Fallback parent is only used for Divisions, not departments.
> Screenshot 2023-11-10 at 9 52 07 AM
> 
> You can notice the arrows on the dogrnut, that is generated by the fallback parent for the division.

OK, would you like me to address that as part of this change or a follow up MR to make sure we are accommodating it? This MR already has a good amount of lines of code changed, even though they're mostly deletes.

AricStewart:
> > You can notice the arrows on the dogrnut, that is generated by the fallback parent for the division.
> 
> OK, would you like me to address that as part of this change or a follow up MR to make sure we are accommodating it? This MR already has a good amount of lines of code changed, even though they're mostly deletes.

If it is not purely cleanup lets do a new MR.  I will approve this one and we can go on from there.

AricStewart (approved):

PR #638, opened 2023-11-13
API: Shift the full DB initialization to the API

AricStewart:
Also removes some test data from the init and moves it to the tests

The 'Historical Placeholder' department is standardized to ID 1, this requires a shift in some of the test that where using 1 as a test Division.


test/DBSeed/Departments.sql
```diff
@@ -1,71 +1,70 @@
-INSERT INTO Departments (DepartmentID, Name, ParentDepartmentID) VALUES (1, 'Activities', 1);
```
rkaiser-dev, line 1: Since we're already in here updating this file anyway, could we update the names of a few of these departments? For example, I think "Smoker's Paradise" was renamed recently, wasn't it?

If we want to not make those updates in this PR, but instead in another, that would be fine too.

rkaiser-dev (approved):
Overall this move makes sense.

Some future thoughts, but would it make sense to simply have all of the SQL executed all the time, so that a brand new instance has _all tables and columns_ regardless of whether those modules are enabled?

If we can get to more dynamically enabled modules, it might make more sense to have all of the DB tables and columns already established up front, so then it's just a matter of "Do I render this?" in the UI. It feels as though the extra care around that probably isn't necessary.

I also believe that it might allow us to approach this with a pattern that allows us to quickly and easily apply new database changes in the future.

PR #639, opened 2023-11-17
Do not rewrite core DB values on init

AricStewart:
If we have existing value for these tables we need them to remain and not be changed as there are existing installs that we will be moving to this new DB code which should not be modified.

### DIFF
--- a/api/src/Core/ModuleDBSchema.php
+++ b/api/src/Core/ModuleDBSchema.php
@@ -34,6 +34,8 @@ abstract class ModuleDBSchema
 
     private $database = null;
 
+    protected $forceOnInstall = true;
+
 
     public function __construct(
         /*.string.*/$field,
@@ -62,7 +64,7 @@ abstract class ModuleDBSchema
         $currentMD5 = md5(json_encode($this->schema));
 
         if ($force || $oldMD5 != $currentMD5) {
-            $force = ($force || $oldMD5 === false);
+            $force = ($force || ($this->forceOnInstall && $oldMD5 === false));
             $this->buildMissingTables($force);
 
             if ($oldMD5 !== false) {
--- a/api/src/Database/DBSchema.php
+++ b/api/src/Database/DBSchema.php
@@ -865,6 +865,7 @@ class DBSchema extends \App\Core\ModuleDBSchema
             $this->index,
             $this->seed
         );
+        $this->forceOnInstall = false;
 
     }
 


rkaiser-dev (commented):
I thought I wouldn't have a great way to test this, but I think I also just figured out why it happened maybe.

I think it's because before, I believe `getRecordedDBSchemaMD5()` was looking for `DBSchemaMD5` and now it's simply `coreDBMD5`, which seems to be what prompted the change for my localhost.

I'm wondering if we should explicitly check for that as a fallback in case `coreDBMD5` isn't present in the DB as an extra cautious check, but it may not be required. Another option might be to ensure that the DB has `coreDBMD5` and it's set to the same value as `DBSchemaMD5` when this goes live.

I see also that other modules have been added as well, separately.

Thoughts on this?

AricStewart:
> I thought I wouldn't have a great way to test this, but I think I also just figured out why it happened maybe.
> 

What I did was initialize the site before this patch (just run any test) and then add the patch and run the test a second time.  It shows how the changes occure.

> I think it's because before, I believe `getRecordedDBSchemaMD5()` was looking for `DBSchemaMD5` and now it's simply `core.DBMD5`, `vol.DBMD5`, etc. which seems to be what prompted the change for my localhost.
> 

Yeah, it was me being clumsy there.  I have made changes so that we do not update a live table from the core table if it already exists and has values.  The are basically no tables that do that except for this "Historical Placeholder" thing in the staff table that just happens to be really important.

> I'm wondering if we should explicitly check for that as a fallback in case `core.DBMD5` isn't present in the DB as an extra cautious check, but it may not be required. Another option might be to ensure that the DB has `core.DBMD5`, `vol.DBMD5`, `staff.DBMD5` and it's set to the same value as `DBSchemaMD5` when this goes live.
> 

Not sure if we need to do that. But also because there are slight changes the DBSchemaMD5 value is going to be off anyway. so even if we do this it will be like a fresh upgrade.

> I see also that other modules have been added as well, separately.
> 
> Thoughts on this?

Good catch. I think I have fixed it.

rkaiser-dev (approved):
> What I did was initialize the site before this patch (just run any test) and then add the patch and run the test a second time. It shows how the changes occure.

I was able to verify this by changing a few things in my pre-existing database. I removed the md5 entries for the newly added `core.`, `vol.`, `staff.` values and updated my Departments table to move the Historical Placeholder Dept to a different number besides 1 to be sure. Loaded the site and confirmed the Departments table was left unchanged. The md5 entries were added back in with current values.

As an aside after thinking about it for a while, I'm not totally sure how I feel about live DB updates such as this. I think for now this is OK, but it would be nice if we had a process to run updates in a more controlled and consistent manner.

PR #641, opened 2023-11-19
DRAFT: Component Library Proof of Concept

rkaiser-dev:
# The Problem

We want a way to quickly implement UI changes without some of the complexity that can exist in implementing those changes. We also want to make sure that our UI has a consistent look and feel across the application, providing a nice user experience throughout.

# Context

I explored several options for how to make general UI improvements in the CIAB site. I looked at the following options:

- [Google's Material Design](https://m3.material.io/) - this would still be a solid choice for us, however we do not use a build process (yet) which would make this a little more challenging. It makes use of Web Components, which are generally supported by all browsers.
- Briefly considered [Material UI](https://mui.com/material-ui/), but it is React only so that was a no-go.
- [Bootstrap](https://getbootstrap.com/docs/5.3/layout/containers/) - has a lot of things that work nicely out of the box. Styling is applied through using typical HTML practices such as defining something like `<div class="the-style">` so it still requires some effort. It's also incompatible with Vue, so it's a no-go.
- [Tailwind CSS](https://tailwindcss.com/) - this is _only_ CSS and also generally recommends use with a build process. Since we are looking for quick implementation for consistent look and feel, this could help in solving some of the issues around theming and consistent look and feel. However, if we wanted to build something like a date picker or color picker, we're still on our own for this.
- [Vuetify](https://vuetifyjs.com/en/) - provides nice, out of the box components for us to leverage, similar to Material Design. This framework is built to work specifically with Vue and gives us the Material Design look and feel, allows theme customization, and more.

Vuetify seems to satisfy several requirements:
- Allows us to quickly implement UI changes through pre-built components
- Allows us to quickly define styles in an application-wide theme.
- All components are rendered in the same way and can be individually customized, or inherit from the application-wide theme.
- Can be used further through our own custom component definitions.
- Can work well with build processes if we leverage that in the future.

For the purposes of this proof of concept I chose Vuetify.

# Scope
This PR includes an example implementation of some changes to our UI using Vuetify.

I chose to implement a very simple theme based on the colors on the Convergence Con site. If desired, we can also somewhat easily incorporate Material UI based Icons, and while I explored this during the example implementation, I noticed that the icons seemed to load in after the page had loaded. With additional effort we may be able to leverage these.

In this example implementation, I chose to define a thing specifically for the Staff Module, because I wanted to limit any visual changes to the ConCom List section of the site _only_. Including the CSS at the root level of the application was causing some unintended visual updates on other parts of the site. 

In practice, it is likely that we will define the theme once, in a single place and then reference that theme when we are applying the Vuetify framework into the application rather than what I have done here as an example.

# Changes Made
- Converted buttons for adding and editing to use Vuetify provided buttons.
- Converted select dropdown for position choices on the sidebar to use Vuetify provided select dropdown

I am including screenshots showing what this looked like before and how it will look after.

As you can see, there was not much that needed to be changed, we were able to simplify the code a little bit and removed some styles that were no longer needed.

This is just a simple example and does not represent fully how we can break out components our own components for further customization and re-use.

PLEASE DO NOT MERGE THIS MR. It would be nice to explore this further and convert more of this to use Vuetify if we feel this is the right path for us.

## Original Buttons and Select Style
ConCom_Buttons_Original
ConCom_Add_Original
ConCom_Modify_Original

## Vuetify-ed Buttons and Select Style
ConCom_Buttons_Vuetify
ConCom_Add_Vuetify
ConCom_Modify_Vuetify

### DIFF
--- a/pages/base/header_start.inc
+++ b/pages/base/header_start.inc
@@ -8,7 +8,8 @@
 
 <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.11.2/css/all.css">
 <script defer src="https://use.fontawesome.com/releases/v5.11.2/js/all.js"></script>
-<script src="https://unpkg.com/vue"></script>
+<script src="https://unpkg.com/vue@3.3.8/dist/vue.global.js"></script>
+<script src="https://unpkg.com/vuetify@3.4.2/dist/vuetify.js"></script>
 <?php
 if (!function_exists('get_console') || get_console() === null) {
     echo "<script src=\"sitesupport/common.js\"></script>\n";
--- a/modules/concom/pages/head.inc
+++ b/modules/concom/pages/head.inc
@@ -5,5 +5,7 @@
 <script src="sitesupport/donut.js"></script>
 <link rel="stylesheet" type="text/css" href="sitesupport/barcode/barcode.css" />
 <link rel="stylesheet" type="text/css" href="theme.css">
+<link rel="stylesheet" href="https://unpkg.com/vuetify@3.4.2/dist/vuetify.css">
+<link href="https://unpkg.com/@mdi/font@5.x/css/materialdesignicons.min.css" rel="stylesheet">
 <script type="module" src="modules/concom/sitesupport/division-parser.js"></script>
 <script type="module" src="modules/concom/sitesupport/concom-list.js"></script>
--- a/modules/concom/scss/concom-list.scss
+++ b/modules/concom/scss/concom-list.scss
@@ -63,14 +63,6 @@
   margin: 1em auto;
 }
 
-.CONCOM-add-member-button {
-  @extend .UI-yellowbutton;
-}
-
-.CONCOM-edit-member-button {
-  @extend .UI-redbutton;
-}
-
 .CONCOM-list-division {
   @extend .UI-container;
   @extend .UI-padding;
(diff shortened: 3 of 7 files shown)


pages/base/header_start.inc
```diff
@@ -8,7 +8,8 @@
 
 <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.11.2/css/all.css">
 <script defer src="https://use.fontawesome.com/releases/v5.11.2/js/all.js"></script>
-<script src="https://unpkg.com/vue"></script>
+<script src="https://unpkg.com/vue@3.3.8/dist/vue.global.js"></script>
```
rkaiser-dev, line 11: Eventually, it would be nice to conditionally include scripts depending on whether we're developing or in production. We can include the base files for us during development, and minified files that have been optimized to load as quickly as possible in production environments.

AricStewart:
We will need to use or replace the `scss/event.scss` functionality.  This is the ability to set the colors based on event.  It is the beginning of a way to skin the site.

This skinning from a user side occurs in `index.php?Function=admin` and `index.php?Function=admin/asset`

The scss defines things like the Primary and secondary event colors that are used on headers and buttons and such. 

I love the idea of this framework, but wonder why you went from red oval buttons to purple square buttons, unless it was just to show what changed.

rkaiser-dev:
> We will need to use or replace the scss/event.scss functionality. This is the ability to set the colors based on event. It is the beginning of a way to skin the site.

If we want to hook it into the pre-configured primary/secondary color scheme from the database we certainly could. I went for simple in this example. So long as we can get the color scheme from an API call, it would be easy to configure.

> I love the idea of this framework, but wonder why you went from red oval buttons to purple square buttons, unless it was just to show what changed.

There were a couple of reasons.

- Demonstrate what we can do with _very little configuration_ on our side. This is the default styling with no additional changes. 
- Simplify the color scheme. Our pages are _way too noisy_ and distracting.

Overall, a few small changes can make the site look and feel _a whole lot better_. We can keep oval buttons if we really want to do that, with minimal effort, but I would advise simplifying the color scheme in general and being consistent with it.

rkaiser-dev:
Closing since this has served its purpose for now. Thanks for the feedback!

PR #645, opened 2023-11-25
API Reorg - Proof of Concept

rkaiser-dev:
## Context
This PR demonstrates a proof of concept to re-organize the API into a more manageable pattern by removing logic from Controllers and pushing that logic down to a Domain layer. Domain in this case refers to the "business logic" required to perform functions in the CIAB Site.

A primary concern that has been addressed with this PR is a major performance issue in which a _significant_ number of SQL queries were being executed while attempting to load the ConCom List page. When requesting and including subresource data, this was resulting in _heavy_ SQL query executions. Subresources in this case may refer to a department within a division, or a staff member within a division or department.

In our current model, including subresources in API calls for Division Staff would result in many queries, up to 2 per staff member alone. On my localhost, with around 2-3 staff members per department, for a single division **at least 99 SQL queries were executed**. I believe this is due in part to a combination of factors including: generic SQL queries that are trying to be all-encompassing and "include" actions that may trigger additional SQL calls when asking for subresource data (i.e., get list of division staff from ConComList, and now fill in member and department data - it seemed to be one at a time for each).

As you review, please ignore any specific implementation typos (such as my use of `/divisions` rather than `/division` for the list of divisions and departments)

## Changes
The changes included represent a shift in design to follow a format that breaks logic up into different logical layers, which helps to keep our Controllers simple and allows us to more easily test our business logic.

In following the new design, I have split out the logic from the Controller into a "Service" and a "Mapper". The Controller will call the appropriate "Service" layer, which may make calls to the "Mapper" layer. These have been added to the Slim Container, so that they can be used as injected dependencies similar to our Database and RBAC code.

The "Service" layer represents "business logic" that may be executed during a request. If we need to format data coming from the database, that would happen here. If we need to format data coming from the UI to match what the database expects before inserting it, that would also happen here. If we need to do any data aggregation or anything else that you might expect from an API, that logic would happen here.

The "Mapper" layer represents calls to the data sources directly. This is sometimes also called a "Repository" layer and we could use that name if we prefer, I was simply following the naming conventions on the Slim documentation. This layer is responsible for fetching data from our Database, from other REST APIs, the filesystem, etc.

I have introduced two new controllers that are specifically dedicated to two functions:
- Get a list of divisions and departments
- Get a list of staff for a given division and its departments

Requesting the staff for a given division in this new implementation has reduced the number of SQL queries down to less than 5 total. Realistically, I think it will be 2-3 in most cases.

API Calls used for Testing:
- `/api/department/`
- `/api/department/{id}/staff?subdepartments=1`
- `/api/divisions`
- `/api/division/{id}/staff` (no need to ask for subdepartments, it is assumed)

While I think the overall concept is demonstrated here, feel free to call out anything that I have missed as far as PHP best practices. I'm still quite unfamiliar with what options are available to get done what I've done here.

## Caveats
- I spent a bit of time investigating our SQL Query usage, and used `EXPLAIN` on our generated queries to figure out if there are additional optimizations we can make there
- I added an index in my SQL database on the `Departments` table (`CREATE INDEX idx_Name ON Departments (Name);`), but I have no idea how we will add this in production currently unless it is just manual.
- I did verify that the data coming from the new and old requests match up for the fields I specifically cared to implement and check against
  - I did this with a JS test suite that I did not include with this PR, comparing against JSON data directly from the APIs I changed
- I did not include PHP tests with these changes
  - The Service Layer can now easily be given a mock implementation that returns consistent data we can test against _without_ needing to connect to a database.
  - We can still test end to end by calling the API directly and connecting to the database as we had done previously
- I only verified against data I had on my local machine
  - One division was used for sampling and proving out the concept. It has 33 staff in the division and related departments.
  - Only one event exists in my database

### DIFF
--- a/api/src/App/Dependencies.php
+++ b/api/src/App/Dependencies.php
@@ -6,6 +6,10 @@
 use Psr\Container\ContainerInterface;
 use App\Handler\ApiError;
 use App\Handler\RBAC;
+use App\Service\DepartmentService;
+use App\Service\EventService;
+use App\Mapper\EventMapper;
+use App\Mapper\StaffMapper;
 use Slim\App;
 
 /* setupAPIDependencies */
@@ -33,6 +37,18 @@ function setupAPIDependencies(App $app, array $settings)
 
     $container['RBAC'] = function (): RBAC {
         return new RBAC;
+    };
+
+    $container['department_service'] = function ($c): DepartmentService {
+        return new DepartmentService(
+            new StaffMapper($c['db'])
+        );
+    };
+
+    $container['event_service'] = function ($c): EventService {
+        return new EventService(
+            new EventMapper($c['db'])
+        );
     };
 
 }
--- a/api/src/App/Routes.php
+++ b/api/src/App/Routes.php
@@ -44,6 +44,13 @@ function setupAPIRoutes(App $app, $authMiddleware)
     )->add(new \App\Middleware\CiabMiddleware($app))->add($authMiddleware);
 
     $app->group(
+        '/divisions',
+        function () use ($app, $authMiddleware) {
+            $app->get('[/]', 'App\Controller\Department\ListDivisions');
+        }
+    )->add(new \App\Middleware\CiabMiddleware($app))->add($authMiddleware);
+
+    $app->group(
         '/deadline',
         function () use ($app, $authMiddleware) {
             $app->get('[/]', 'App\Controller\Deadline\ListMemberDeadlines');
--- a/api/src/Controller/Department/ListDivisions.php
+++ b/api/src/Controller/Department/ListDivisions.php
@@ -1,0 +1,70 @@
+<?php declare(strict_types=1);
+
+/**
+ * @OA\Get(
+ *     tags={"departments"},
+ *     path="/divisions",
+ *     summary="Lists all divisions and their departments",
+ *     @OA\Response(
+ *         response=200,
+ *         description="OK",
+ *         @OA\JsonContent(
+ *             ref="#/components/schemas/division_list"
+ *         )
+ *     ),
+ *     @OA\Response(
+ *         response=401,
+ *         ref="#/components/responses/401"
+ *     ),
+ *     security={{"ciab_auth":{}}}
+ * )
+ *
+ * @OA\Schema(
+ *    schema="division_list",
+ *    allOf = {
+ *        @OA\Schema(ref="#/components/schemas/resource_list")
+ *    },
+ *    @OA\Property(
+ *        property="type",
+ *        type="string",
+ *        enum={"division_list"}
+ *    ),
+ *    @OA\Property(
+ *        property="data",
+ *        type="array",
+ *        description="List of divisions and departments",
+ *        @OA\Items(
+ *            ref="#/components/schemas/department"
+ *        )
+ *    )
+ * )
+ */
+namespace App\Controller\Department;
+
+use Slim\Container;
+use Slim\Http\Request;
+use Slim\Http\Response;
+
+class ListDivisions
+{
+
+    protected $departmentService;
+
+
+    public function __construct(Container $container)
+    {
+        $this->departmentService = $container->get('department_service');
+
+    }
+
+
+    public function __invoke(Request $request, Response $response, $args)
+    {
+        $divisions = $this->departmentService->getStaffDivisions();
+        return $response->withJson($divisions, 200);
+
+    }
+
+
+    /* End ListDivisions */
+}
--- a/api/src/Mapper/EventMapper.php
+++ b/api/src/Mapper/EventMapper.php
@@ -1,0 +1,46 @@
+<?php declare(strict_types=1);
+
+namespace App\Mapper;
+
+use Atlas\Query\Select;
+
+class EventMapper
+{
+
+    protected $db;
+
+
+    public function __construct($db)
+    {
+        $this->db = $db;
+
+    }
+
+
+    public function getMostRecentEvent()
+    {
+        $select = Select::new($this->db);
+        return $select
+            ->columns('EventID', 'AnnualCycleID', 'DateFrom', 'DateTo', 'EventName')
+            ->from('Events')
+            ->where('Events.DateTo >= NOW()')
+            ->orderBy('Events.DateFrom ASC LIMIT 1')
+            ->fetchOne();
+
+    }
+
+
+    public function getLastActiveEvent()
+    {
+        $select = Select::new($this->db);
+        return $select
+            ->columns('EventID', 'AnnualCycleID', 'DateFrom', 'DateTo', 'EventName')
+            ->from('Events')
+            ->orderBy('Events.EventID DESC LIMIT 1')
+            ->fetchOne();
+
+    }
+
+    
+    /* End EventMapper */
+}
--- a/api/src/Mapper/StaffMapper.php
+++ b/api/src/Mapper/StaffMapper.php
@@ -1,0 +1,101 @@
+<?php declare(strict_types=1);
+
+namespace App\Mapper;
+
+use Atlas\Query\Select;
+
+class StaffMapper
+{
+
+    protected $db;
+
+
+    public function __construct($db)
+    {
+        $this->db = $db;
+
+    }
+
+
+    public function getStaffDivisions()
+    {
+        $select = Select::new($this->db);
+        $select->columns('DepartmentID as id', 'ParentDepartmentID as parent', 'Name as name', 'FallbackID as fallback')
+            ->columns(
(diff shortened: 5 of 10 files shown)


AricStewart:
Ok I am going to read more and again more closely but my gut is that I like this.  I am seeing where you are going and that makes sense.  I am not sure if I see how this is going to get us our improvement or solve the performance hit for including resources. But maybe you can explain that. 

Also part of the structure is that things in the `Modules` directory are able to be turned on and off. So you can run and event without staff turned on, and all the things about staff is contained in that one module.  Same for Volunteers, or Registration or whatever we write that may be turned on and off.

If you could work this example back into the Module that would be helpful for me. 

Finally looking at the `StaffMapper`  I understand but am also not completely sold.  What you are doing is eliminating the include logic, you eliminate the connection between a staff in a position identified by an `AccountID` and a meber identified by the same `AccountID`  and that creates a lot of duplication of queries...

Just brainstorming I wonder if we have mapper objects have an interface that returns the `Select` statement. So once can build single SQL statements from all the various clauses gathered by the various object that are being included in the return. 

That make sense? If you are foggy I will try to write up a mock-up tomorrow.

rkaiser-dev:
> If you could work this example back into the Module that would be helpful for me.

I think in those cases where we disable the module, the code is still hanging around though, right? In particular I notice that Departments are not currently part of the Staff module, which is how this ended up where it did. If needed you could always put the module specific services under `Modules/{theModule}/Service` or `Modules/{theModule}/Mapper` or whatever.

Then, instead of dropping the service or mapper dependency in the main file for `App/Dependencies.php` you could put it instead in `Module/{theModule}/App/Dependencies.php`. That wouldn't be a problem. Add the function name to the module `Settings.php` code to call the dependency setup in the module and good to go.

>  What you are doing is eliminating the include logic, you eliminate the connection between a staff in a position identified by an AccountID and a meber identified by the same AccountID and that creates a lot of duplication of queries...

I'm not sure I entirely follow this. I don't think I've removed the connection between Staff and Member identified by the same `AccountID`. In the example below, you can see that I am getting the necessary data for the UI to build everything on the ConCom List page with just a single SQL query (not including the query to find the current event).

An example of the output from the old API call and the new one are included below. Please ignore that the `type` in the entry is not defined. That would be a pretty easy add, but I'm not sure if we even use that. I certainly didn't in my UI code.

`/api/department/{id}/staff?subdepartments=1`

```
  {
    "type": "staff_entry",
    "id": "1",
    "member": {
      "type": "member",
      "id": "1000",
      "legal_first_name": "Test",
      "legal_last_name": "Admin",
      "middle_name": "A",
      "suffix": null,
      "email": "test-admin@example.com",
      "email2": null,
      "email3": null,
      "phone": "555-555-5555",
      "phone2": null,
      "address_line1": "1234",
      "address_line2": null,
      "city": null,
      "state": "MN",
      "zip_code": null,
      "zip_plus4": null,
      "country": "United States of America",
      "province": null,
      "preferred_first_name": null,
      "preferred_last_name": null,
      "deceased": null,
      "do_not_contact": null,
      "email_optout": null,
      "birthdate": null,
      "gender": null,
      "concom_display_phone": null,
      "dependent_on": null,
      "pronouns": "He/Him",
      "first_name": "Test",
      "last_name": "Admin",
      "duplicates": null
    },
    "department": {
      "id": "2",
      "parent": null,
      "name": "Activities",
      "fallback": null,
      "child_count": "13",
      "email": [
        "activities@test-con.org"
      ],
      "type": "department"
    },
    "positionId": "1",
    "note": "Test Note",
    "position": "Head"
  }
```

`/api/divisions/{id}/staff`

```
  {
    "id": "1",
    "member": {
      "id": "1000",
      "first_name": "Test",
      "last_name": "Admin",
      "email": "test-admin@example.com",
      "pronouns": "He/Him"
    },
    "department": {
      "id": "2",
      "name": "Activities"
    },
    "note": "Test Note",
    "position": "Head"
  }
```

In the new call, if the department has a parent then that would change `department` to look like:

```
"department": {
  "id": "105",
  "name": "Book Swap",
  "parent": {
    "id": "2",
    "name": "Activities"
  }
}
```

As for duplicated queries, I don't think that's necessarily the case. Conditionally building queries can make things a bit difficult to follow. I think we can apply _some_ level of abstraction or DRY to the queries to make sure they're not _all unique_, but I do think we should be careful not to get _too abstract_ or _too conditional_ with them.

For example, if we want to get all staff positions for a member, we could somewhat easily extract common parts of the query included in this example.

I believe, just by looking at the output on the `/member/{id}/staff_membership`, that the following could be pulled out into a private function.

```
private function getStaffMemberSelect()
{
        $select = Select::new($this->db);
        $select
            ->columns('Staff.ListRecordID as id', 'Staff.AccountID as member', 'Staff.DepartmentID as department')
            ->columns('COALESCE(Staff.Note, "") as note')
            ->columns(
                $select->subSelect()
                    ->columns('Name')
                    ->from('ConComPositions')
                    ->where('PositionID = Staff.PositionID')
                    ->as('position')->getStatement()
            )
            ->columns('Depts.Name as departmentName', 'Depts.ParentDepartmentID as parentId', 'Divs.Name as parentName')
            ->columns('(CASE WHEN Members.PreferredFirstName IS NOT NULL THEN Members.PreferredFirstName ELSE Members.FirstName END) as first_name')
            ->columns('(CASE WHEN Members.PreferredLastName IS NOT NULL THEN Members.PreferredLastName ELSE Members.LastName END) as last_name')
            ->columns('Members.Email as email', 'Members.Pronouns as pronouns')
            ->from('ConComList as Staff')
            ->join('INNER', 'Members', 'Staff.AccountID = Members.AccountID')
            ->join('INNER', 'Departments as Depts', 'Staff.DepartmentID = Depts.DepartmentID')
            ->join('INNER', 'Departments as Divs', 'Depts.ParentDepartmentID = Divs.DepartmentID');
}
```

This will give us the SQL we need to pull a standard selection of data from the staff member. I'm fairly certain we could add columns on top of this if we wanted to add more data. Then, we simply use the `WHERE` conditions that we want to apply here. The overall function might look like this:

```
// StaffMapper.php

public function getStaffMember($eventId, $memberId)
{
        $select = $this->getStaffMemberSelect();
        $select->where('Staff.EventID = ', $eventId)
              ->where('Staff.AccountID = ', $memberId);

        return $select->fetchAll();
}
```

I'm open to further suggestion here if I've misunderstood, but I think by using _very targeted_ SQL, we can make the performance that much better.

And this can solve the performance issue with includes by using specifically targeted SQL for the cases we care about. I can think of a couple of examples where using generic SQL in general is hurting us a bit. In particular, when searching for division staff members, I think we first grab the division and all departments, and then we get the staff and iterate over them individually - populating member details and department details, something that produces N+2 queries where N is the number of staff. By instead crafting specifically targeted SQL knowing what I wanted, I was able to get this down to just 1 query.

There are also a couple of examples that stood out, and while not specifically related to includes, I noticed some unnecessary queries and `WHERE` clauses being executed. Getting Current Event, and getting Current Member data.

For Current Event, I know that I will need to determine what event that is. But I also know that I can just grab the entire Event row that comes back from the DB and shape it the way we need it. Right now, if we want to get the Current Event we are first looking up the Event ID in the DB, and then going back and making a second call to get the rest of the row data using that ID. That's two calls that we can condense into a single call. In my example code with `EventMapper.php` I am duplicating some of the `$select` such as the columns and from which table, but we could extract that into a private function too. Then just change our conditions. We could add a function specifically targeting cases where we know the ID, and then have a separate function in `EventService` to handle that case.

This way, if our Controller is answering `/event/{id}` it can call `EventService->getById($eventId)` and the EventService is calling `EventMapper->getById($eventId)`. It targets specifically one column without needing to do any extra checking first, and it gets us the data we want.

For Current Member, the lookup code goes into a generic function related to member lookup that covers multiple cases, but in our case it ends up trying to look up the user by ID and Email. But we already have the member ID from the OAuth token. We wouldn't even need to consider email at all, since we have the ID. It simplifies the query immensely because we should be able to simply look up by `AccountID`. Then we can break up the other "find member by" things into separate queries that target what we want. If we wanted to combine a couple, we could make a mapper function for that.

Hope this helps!

rkaiser-dev:
Also, if there is still concern about how we're performing the SQL Queries in terms of duplication, perhaps an ORM library would help us out here? We wouldn't need to roll our own and it would simplify our implementation a bit. It probably isn't necessary, but would make things easier for folks who aren't as familiar with how SQL works. Many also allow you access to make custom queries as well, if you find that some queries are not particularly performant.

AricStewart:
Response sent in e-mail for now.

rkaiser-dev:
Closing this MR in favor of more elegant approach.

PR #647, opened 2023-11-27
Convert Department List API to Controller -> Service -> Repo pattern

rkaiser-dev:
This PR includes changes necessary to convert from the current inherited Controller model to the Controller -> Service -> Repository pattern.

The changes were made as a first step toward improving overall API performance for this API. I left the `buildResource` function alone for now. Once we are confident with these changes I would recommend removing `buildResource` as a fast-follow. We can also remove the `extends BaseDepartment` at that time, as it will no longer be needed.

When fetching department details I made the judgment call to supply only an ID and Name for the Parent and Fallback Department references. I figured that the data is already available in the response if more than that needs to be used.

I did not add tests, but most of that stems from lack of familiarity with testing approaches in PHP. I can spend some time on it if we'd like to get them added and include them in a follow-up MR.

With just a few sample calls, I was seeing that the response time for this endpoint was originally around 300ms on my local machine. I believe I had configured some fallback departments on my divisions.

After making these changes, the new request was typically completing around 80ms.

### DIFF
--- a/api/src/Controller/Department/ListDepartments.php
+++ b/api/src/Controller/Department/ListDepartments.php
@@ -34,16 +34,31 @@
 
 namespace App\Controller\Department;
 
+use Slim\Container;
 use Slim\Http\Request;
 use Slim\Http\Response;
+
+use App\Service\DepartmentService;
 
 class ListDepartments extends BaseDepartment
 {
 
+    /**
+     * @var DepartmentService
+     */
+    protected $departmentService;
+
+
+    public function __construct(Container $container)
+    {
+        parent::__construct($container);
+        $this->departmentService = $container->get("DepartmentService");
+        $this->includes = null;
+    }
 
     public function buildResource(Request $request, Response $response, $params): array
     {
-        $output = $this->getDepartment(null);
+        $output = $this->departmentService->getAllDepartments();
         return [
         \App\Controller\BaseController::LIST_TYPE,
         $output,
--- a/api/src/Repository/DepartmentRepository.php
+++ b/api/src/Repository/DepartmentRepository.php
@@ -1,0 +1,63 @@
+<?php declare(strict_types=1);
+
+namespace App\Repository;
+
+use Atlas\Query\Select;
+
+class DepartmentRepository
+{
+
+    protected $db;
+
+
+    public function __construct($db)
+    {
+        $this->db = $db;
+
+    }
+
+
+    public function findAll()
+    {
+        $select = Select::new($this->db);
+        $select->columns(
+            'depts.DepartmentID',
+            'depts.Name',
+            'parentDepts.DepartmentID as ParentID',
+            'parentDepts.Name as ParentName',
+            'fallbackDepts.DepartmentID as FallbackID',
+            'fallbackDepts.Name as FallbackName'
+        )
+            ->columns(
+                $select->subSelect()
+                    ->columns('COUNT(DepartmentID)')
+                    ->from('Departments as childDepts')
+                    ->where('childDepts.ParentDepartmentID = depts.DepartmentID')
+                    ->andWhere('Name != "Historical Placeholder"')
+                    ->as('ChildCount')
+                    ->getStatement()
+            )
+            ->columns(
+                $select->subSelect()
+                    ->columns('GROUP_CONCAT(Email)')
+                    ->from('EMails')
+                    ->where('DepartmentID = depts.DepartmentID')
+                    ->as('Email')
+                    ->getStatement()
+            )
+            ->from('Departments as depts')
+            ->join('LEFT', 'Departments as parentDepts', 'parentDepts.DepartmentID = depts.ParentDepartmentID')
+            ->join('LEFT', 'Departments as fallbackDepts', 'fallbackDepts.DepartmentID = depts.FallbackID');
+
+        $historicalPlaceholders = $select->subSelect()->columns('DepartmentID')->from('Departments')->where('Name = "Historical Placeholder"');
+
+        $select->where('depts.DepartmentID NOT IN ', $historicalPlaceholders)
+            ->where('depts.ParentDepartmentID NOT IN ', $historicalPlaceholders);
+
+        return $select->fetchAll();
+
+    }
+
+
+    /* End DepartmentRepository */
+}
--- a/api/src/Service/DepartmentService.php
+++ b/api/src/Service/DepartmentService.php
@@ -1,0 +1,66 @@
+<?php declare(strict_types=1);
+
+namespace App\Service;
+
+use App\Repository\DepartmentRepository;
+
+class DepartmentService
+{
+
+  /**
+   * @var DepartmentRepository
+   */
+    protected $departmentRepository;
+
+
+    public function __construct(DepartmentRepository $departmentRepository)
+    {
+        $this->departmentRepository = $departmentRepository;
+
+    }
+
+
+    public function getAllDepartments()
+    {
+        $result = $this->departmentRepository->findAll();
+
+        $formatted = [];
+        foreach ($result as $value) {
+            $output = [];
+            $output["id"] = $value["DepartmentID"];
+            $output["name"] = $value["Name"];
+            $output["child_count"] = $value["ChildCount"];
+            $output["type"] = "department";
+
+            if ($value["Email"]) {
+                $output["email"] = explode(',', $value["Email"]);
+            } else {
+                $output["email"] = [];
+            }
+
+            if ($value["ParentID"] && $value["ParentID"] != $value["DepartmentID"]) {
+                $output["parent"]["id"] = $value["ParentID"];
+                $output["parent"]["name"] = $value["ParentName"];
+                $output["parent"]["type"] = "department";
+            } else {
+                $output["parent"] = null;
+            }
+
+            if ($value["FallbackID"]) {
+                $output["fallback"]["id"] = $value["FallbackID"];
+                $output["fallback"]["name"] = $value["FallbackName"];
+                $output["fallback"]["type"] = "department";
+            } else {
+                $output["fallback"] = null;
+            }
+
+            $formatted[] = $output;
+        }
+
+        return $formatted;
+
+    }
+
+
+    /* End DepartmentService */
+}
--- a/api/src/App/Dependencies.php
+++ b/api/src/App/Dependencies.php
@@ -6,6 +6,8 @@
 use Psr\Container\ContainerInterface;
 use App\Handler\ApiError;
 use App\Handler\RBAC;
+use App\Repository\DepartmentRepository;
+use App\Service\DepartmentService;
 use Slim\App;
 
 /* setupAPIDependencies */
@@ -33,6 +35,12 @@ function setupAPIDependencies(App $app, array $settings)
 
     $container['RBAC'] = function (): RBAC {
         return new RBAC;
+    };
+
+    $container['DepartmentService'] = function ($c): DepartmentService {
+        return new DepartmentService(
+            new DepartmentRepository($c['db'])
+        );
     };
 
 }


api/src/Controller/Department/ListDepartments.php
```diff
@@ -34,13 +34,38 @@
...
+    public function __invoke(Request $request, Response $response, $params)
+    {
+        $output["type"] = "department_list";
+        $output["data"] = $this->departmentService->getAllDepartments();
+        return $response->withJson($output, 200);
+
+    }
+
 
+    // This function ignored with __construct and __invoke uncommented. Left here for now for reference.
```
AricStewart, line 68: There is a LOT of backing work that is now being thrown away here that I have not been talked into yet.  If you look at the `BaseController` invoke you can see lots of error handling and automatic processing of the Json and includes all in one place.  You are abandoning all of that structure. 

Why are we not for a first step putting your new code into buildResource and returning the structure expected there.  That would make me much more comfortable with this change as it would be more incremental.

api/src/Repository/DepartmentRepository.php
```diff
@@ -0,0 +1,63 @@
...
+            ->from('Departments as depts')
+            ->join('LEFT', 'Departments as parentDepts', 'parentDepts.DepartmentID = depts.ParentDepartmentID')
+            ->join('LEFT', 'Departments as fallbackDepts', 'fallbackDepts.DepartmentID = depts.FallbackID');
+
+        $historicalPlaceholders = $select->subSelect()->columns('DepartmentID')->from('Departments')->where('Name = "Historical Placeholder"');
+
+        $select->where('depts.DepartmentID NOT IN ', $historicalPlaceholders)
+            ->where('depts.ParentDepartmentID NOT IN ', $historicalPlaceholders);
+
+        return $select->fetchAll();
```
AricStewart, line 57: So there is a pretty powerful system I build around the `columnsToAttributes` member.  I can understand wanting it pull it out into a `BaseRepository` or something out of the overloaded `BaseController`. I also notice it is not being used in the current Department code, which is disappointing, maybe because the merging of all the tables is being complicated, but it will still work, you can specified fields like `'<table>.<field>' => '<output>',` so `depts.DepartmentID =-> 'id',` would work and I feel is the better way to go.

Look at `Member/BaseMember.php` and see the definition for `protected static $columnsToAttributes` 
Then look at `Member/FindMember.php` and you can see `->columns(...BaseMember::selectMapping())` which programmatically builds all the Database field to resource value mapping. 

I see that a lot of that mapping has moved into the `service` below but it feels like there should be a cleaner way to express it, maybe still using a method like `columnsToAttributes`?  The reason SO many of the DB fields are different that the member fields is because the DB fields where inherited and do not fit a clean RESTful structure as I understood it.

api/src/Service/DepartmentService.php
```diff
@@ -0,0 +1,66 @@
...
+            if ($value["FallbackID"]) {
+                $output["fallback"]["id"] = $value["FallbackID"];
+                $output["fallback"]["name"] = $value["FallbackName"];
+                $output["fallback"]["type"] = "department";
+            } else {
+                $output["fallback"] = null;
+            }
+
+            $formatted[] = $output;
+        }
```
AricStewart, line 58: See above

AricStewart (changes requested):

api/src/Controller/Department/ListDepartments.php
rkaiser-dev: >  If you look at the BaseController invoke you can see lots of error handling and automatic processing of the Json and includes all in one place. You are abandoning all of that structure.

> Why are we not for a first step putting your new code into buildResource and returning the structure expected there. That would make me much more comfortable with this change as it would be more incremental.

I'd like to answer both of these at the same time, because I think it will help illustrate the decision I made here to override invoke.

For error handling, I can see the reasoning in keeping this behavior. I think that this is something that would be best handled through the use of a middleware rather than an abstract class. It will help us move away from a model of inheritance in favor of one that leverages composition. In the case of this specific call, the only relevant HTTP Statuses are 200 (we got the data), 404 (we didn't have any data), and 500 (we goofed). I can certainly carry those over into this controller for now.

I'm not sure what you mean by automatic JSON processing exactly. I do see some logic that handles query parameters, which I think is valid. Is that what you're pointing out?

Can you help me understand why those were added initially, and whether we would need to keep them for this call? I know the obvious answer is so that we can paginate through data in case of _many results_, but for departments I didn't anticipate that there would be so many that it would be a problem.

As for skipping include logic, is there a concern about that here? I believe we are including all of the relevant information for Divisions and Departments with this request, including enough about parent and fallback departments that API consumers can access what they need from the rest of the response. I'd be willing to concede that we are not returning the full object for parent and fallback department, but again it felt like it was not required because that data exists elsewhere in the response.

Regarding putting new code into `buildResource`, I think that largely most of the underlying code wasn't required to fetch the data for this endpoint. I will concede that battle-tested error handling and query parameter processing would be nice to have. However, I would need to eliminate the includes altogether because I would not want those to be run, ever, for this endpoint. It might be that I could replace includes in the `ListDepartments` controller with an empty array.

For better or worse, it felt cleaner to have a fresh path for this code. Most of the underlying functionality has been replaced, so I didn't want to add confusion by leveraging the abstract function because a lot of the request processing didn't feel applicable. It felt best to start pulling the bandage off, if that makes sense.

I agree that it might be more incremental to use `buildResource`, but a number of parameters would've been flat out ignored anyway unless I implemented them in the service and repository layer, which could be done and isn't impossible, but based on the performance it seemed fine if we didn't.

api/src/Repository/DepartmentRepository.php
rkaiser-dev: > So there is a pretty powerful system I build around the `columnsToAttributes` member. I can understand wanting it pull it out into a `BaseRepository` or something out of the overloaded `BaseController`.

I think `columnsToAttributes` could be useful in the future, but as we are implementing just a single repository here it did not feel necessary to make that abstraction at this time. While we can anticipate that there will be things in the future to work through, I think there are better options than to rely on another `Base` class and inheritance when we really want composition.

Additionally, it is not typical that the repository layer is doing any of this kind of mapping. That is typically reserved for the service layer in this pattern. The repository is typically responsible for making queries to the database and nothing more.

Perhaps another approach would be to make entity classes as a middle ground (which are also not using inheritance). In the constructor you can have it accept a row and it manages the mapping from SQL row to JSON on its own. We could also leverage an implementation for `equals` if we needed to do so, and if really needed, maybe some getters on the attributes.

I'd like to avoid adding complexity to our SQL queries because we should be able to pretty quickly put together what call is being made to the database. While `columnsToAttributes` wouldn't necessarily detract from this, I think there is a time and place where we could better implement this because it's very likely that we will want repository calls that _are not_ going to produce API output.

We also don't even have to map from Database column names to API friendly names (because I've seen `keyLikeThis`, `key_like_this` and `KeyLikeThis` come back from APIs. However it might be nice to hide the implementation (a little) or else we risk tightly coupling our UI code. I would make a strong case that this implementation should be hidden in the service layer, or with entities that represent the database objects, similar to how you'd see for an ORM library.

api/src/Controller/Department/ListDepartments.php
AricStewart: Just as an overview, I really like this design and path forward but I am being very incremental and cautious at our implementation.  There actually has been a LOT of design and thought put into this and while the structure is a bit of a mess I want to try to make changes more incrementally and with the same or more thought and planning.

> > If you look at the BaseController invoke you can see lots of error handling and automatic processing of the Json and includes all in one place. You are abandoning all of that structure.
> 
> > Why are we not for a first step putting your new code into buildResource and returning the structure expected there. That would make me much more comfortable with this change as it would be more incremental.
> 
> I'd like to answer both of these at the same time, because I think it will help illustrate the decision I made here to override invoke.
> 
> For error handling, I can see the reasoning in keeping this behavior. I think that this is something that would be best handled through the use of a middleware rather than an abstract class. It will help us move away from a model of inheritance in favor of one that leverages composition. In the case of this specific call, the only relevant HTTP Statuses are 200 (we got the data), 404 (we didn't have any data), and 500 (we goofed). I can certainly carry those over into this controller for now.
> 

Could you help me see the middleware approach here?  Maybe we implement this middleware first because then that logic can be removed from the existing code path. 

> I'm not sure what you mean by automatic JSON processing exactly. I do see some logic that handles query parameters, which I think is valid. Is that what you're pointing out?
> 
> Can you help me understand why those were added initially, and whether we would need to keep them for this call? I know the obvious answer is so that we can paginate through data in case of _many results_, but for departments I didn't anticipate that there would be so many that it would be a problem.
> 
> As for skipping include logic, is there a concern about that here? I believe we are including all of the relevant information for Divisions and Departments with this request, including enough about parent and fallback departments that API consumers can access what they need from the rest of the response. I'd be willing to concede that we are not returning the full object for parent and fallback department, but again it felt like it was not required because that data exists elsewhere in the response.
> 

I am looking at the `__invoke` implementation in 'BaseController'. The base goal was to have very consistent reply construction such as ensuring that ever returned resource has an `id` and a `type` and that lists are consistently paged.  Also the goal was to remove as much of all that from the individual endpoint implementations as possible. 

I need to dig more into the include logic because you are insistent that the return is the same and I am not seeing that in the code so maybe I am missing something.  I would really like this change be unchanging to the schema. If we want to change the schema for departments or staff that is fine but that is not part of a backend change like this. It should be done independently. 

> Regarding putting new code into `buildResource`, I think that largely most of the underlying code wasn't required to fetch the data for this endpoint. I will concede that battle-tested error handling and query parameter processing would be nice to have. However, I would need to eliminate the includes altogether because I would not want those to be run, ever, for this endpoint. It might be that I could replace includes in the `ListDepartments` controller with an empty array.
> 
> For better or worse, it felt cleaner to have a fresh path for this code. Most of the underlying functionality has been replaced, so I didn't want to add confusion by leveraging the abstract function because a lot of the request processing didn't feel applicable. It felt best to start pulling the bandage off, if that makes sense.
> 
> I agree that it might be more incremental to use `buildResource`, but a number of parameters would've been flat out ignored anyway unless I implemented them in the service and repository layer, which could be done and isn't impossible, but based on the performance it seemed fine if we didn't.

I am torn here because I understand you sentiment, yet when making changes like this to a core part of the system I am far less of a "tearing bandage" person and more of a "progressive incremental change" person.  Like I said above I would really want this change to be schema identical. 

Maybe this is just because I am the author here but even if not a scrap of the old code exists after this restructure I would be far more comfortable taking this in smaller steps replacing it bit by bit than a whole hog reimplementation from the ground up. 

I would be willing to be more relaxed about this for something like the `Staff` submodule, but when we are doing `Cycles` and `Events` and `Departments` we are touch things that touch many other things and we would have to be sure everything still works together at each step, and not just because it is still using the old code in parallel.

api/src/Repository/DepartmentRepository.php
AricStewart: > > So there is a pretty powerful system I build around the `columnsToAttributes` member. I can understand wanting it pull it out into a `BaseRepository` or something out of the overloaded `BaseController`.
> 
> I think `columnsToAttributes` could be useful in the future, but as we are implementing just a single repository here it did not feel necessary to make that abstraction at this time. While we can anticipate that there will be things in the future to work through, I think there are better options than to rely on another `Base` class and inheritance when we really want composition.
> 
> Additionally, it is not typical that the repository layer is doing any of this kind of mapping. That is typically reserved for the service layer in this pattern. The repository is typically responsible for making queries to the database and nothing more.
> 
> Perhaps another approach would be to make entity classes as a middle ground (which are also not using inheritance). In the constructor you can have it accept a row and it manages the mapping from SQL row to JSON on its own. We could also leverage an implementation for `equals` if we needed to do so, and if really needed, maybe some getters on the attributes.
> 
> I'd like to avoid adding complexity to our SQL queries because we should be able to pretty quickly put together what call is being made to the database. While `columnsToAttributes` wouldn't necessarily detract from this, I think there is a time and place where we could better implement this because it's very likely that we will want repository calls that _are not_ going to produce API output.
> 
> We also don't even have to map from Database column names to API friendly names (because I've seen `keyLikeThis`, `key_like_this` and `KeyLikeThis` come back from APIs. However it might be nice to hide the implementation (a little) or else we risk tightly coupling our UI code. I would make a strong case that this implementation should be hidden in the service layer, or with entities that represent the database objects, similar to how you'd see for an ORM library.

Yes I would like to "hide the implementation". 
I can see the argument for the mapping changing location and no longer having it part of the SQL query. I just would like to see a clean generalized system for it instead of each mapping layer having to have a messy large `output[X] = value[Y]` block.

api/src/Controller/Department/ListDepartments.php
rkaiser-dev: > I want to try to make changes more incrementally and with the same or more thought and planning.

I think that's fair. The way I proposed these changes by completely bypassing the stuff from `BaseController` is a bit drastic, compared to a more incremental approach. I think it can probably be achieved incrementally by responding with the data I've got from the new pattern in `buildResource`, and make sure I prevent any `includes` from running unless we really want to run them. 

My preference for this call is to avoid them, for two reasons:
- We would lose out on the performance benefits of this call if we kept them, even with re-organizing the code a bit.
- I think the only code currently using this endpoint is on the ConCom List page. This isn't removing the call in the `BaseController` so other areas won't be impacted by this until we move them to the new pattern
- The data is about as schema-identical as we can reasonably get

I have attached sample responses for the original and the updated request.
[original_sample.json](https://github.com/CON-In-A-Box/CIAB-Portal/files/13506729/original_sample.json)
[updated_sample.json](https://github.com/CON-In-A-Box/CIAB-Portal/files/13506730/updated_sample.json)

In both we have `id`, `name`, `child_count`, `email`. 

In the old, `parent` also includes _the entire parent object_ which is present earlier in the response output. If that parent had `fallback` values, you'd see the entire chain of `fallback` departments in full.

In the new, `parent` simply contains `id` and `name`. `fallback` in the new is similarly just `id` and `name`.

In this specific API request, we will always have a full list of departments. We always populate parent and fallback ID and Name if present. It felt redundant to walk any further down the list, if that makes sense. If we look at a department and it has a parent, we have the info we need to find the parent department if we want more detail. In that sense, we aren't missing any data.

I deliberately chose to ignore the case where we look up a single department by ID. In a call like that, we could certainly provide a bit more detail about the parent or fallback, but in a request that gets us a list it felt less applicable.

I think for the other areas you mention that I agree, it will be good to make sure they work at every step. It seems that one of the primary concerns (unless I've misunderstood) is to make sure the UI has everything it needs from an API call, but I also want to make sure we don't conflate that with "API must return all DB fields exactly", either, because it may not be required.

api/src/Repository/DepartmentRepository.php
rkaiser-dev: I'm in agreement, having a clean system for managing this is something we can work toward. I think I see a way we can incorporate this without having to use the functions that are directly on `BaseController` by extracting those into their own function. We might even be able to break them out and then let `BaseController` call them.

What can we do to handle nested objects? For example if we want to take `ParentID` and `ParentName` and put them in an output that is like so: `{ "parent": { "id": "123", "name": "The Name" } }`?

AricStewart: > I'm in agreement, having a clean system for managing this is something we can work toward. I think I see a way we can incorporate this without having to use the functions that are directly on `BaseController` by extracting those into their own function. We might even be able to break them out and then let `BaseController` call them.
> 

I do not want to be taking steps back to take steps forward here.  Instead I would like sideways but non-regressive steps. 

> What can we do to handle nested objects? For example if we want to take `ParentID` and `ParentName` and put them in an output that is like so: `{ "parent": { "id": "123", "name": "The Name" } }`?

This is a bit of an extreme example.  But live..
When pulling a staff list with `api/department/2/staff` we get a response that is not unlike this:

```
{
    "type": "staff_list",
    "data": [
        {
            "type": "staff_entry",
            "id": "1022",
            "member": {
                "type": "member",
                "id": "1000",
                "legal_first_name": "Odin",
                "legal_last_name": "Allfather",
                "middle_name": null,
                "suffix": null,
                "email": "allfather@oneeye.com",
                "email2": null,
                "email3": null,
                "phone": null,
                "phone2": null,
                "address_line1": null,
                "address_line2": null,
                "city": null,
                "state": null,
                "zip_code": null,
                "zip_plus4": null,
                "country": null,
                "province": null,
                "preferred_first_name": null,
                "preferred_last_name": null,
                "deceased": null,
                "do_not_contact": null,
                "email_optout": null,
                "birthdate": null,
                "gender": "Allfather",
                "concom_display_phone": null,
                "dependent_on": null,
                "pronouns": null,
                "first_name": "Odin",
                "last_name": "Allfather",
                "duplicates": null
            },
            "department": {
                "id": "2",
                "parent": null,
                "name": "Activities",
                "fallback": {
                    "id": "8",
                    "parent": null,
                    "name": "Systems",
                    "fallback": {
                        "id": "7",
                        "parent": null,
                        "name": "Productions",
                        "fallback": {
                            "id": "6",
                            "parent": null,
                            "name": "Hospitality",
                            "fallback": {
                                "id": "5",
                                "parent": null,
                                "name": "Facilities",
                                "fallback": {
                                    "id": "4",
                                    "parent": null,
                                    "name": "External Relations and Communications",
                                    "fallback": {
                                        "id": "3",
                                        "parent": null,
                                        "name": "Administration",
                                        "fallback": "2",
                                        "child_count": "8",
                                        "email": [],
                                        "type": "department"
                                    },
                                    "child_count": "9",
                                    "email": [],
                                    "type": "department"
                                },
                                "child_count": "10",
                                "email": [],
                                "type": "department"
                            },
                            "child_count": "9",
                            "email": [],
                            "type": "department"
                        },
                        "child_count": "9",
                        "email": [],
                        "type": "department"
                    },
                    "child_count": "2",
                    "email": [],
                    "type": "department"
                },
                "child_count": "9",
                "email": [],
                "type": "department"
            },
            "positionId": "1",
            "note": "",
            "position": "Head"
        }
    ]
}
```

This is because the `staff_list` resource is a list of `staff_entry` and each `staff_entry` has a `member` field which expands into a `member` resource and a `department` field that expands into a `department` resource and so on.  This accordion of inclusions ensures that all the data that could be needed is present.  This is a bit of a mess because with the fallbacks we end up getting the whole fallback structure as well as other data we do not need.  This is likely part of the pain is that all these resources are getting queried over and over.  If we have a lot of people in this department then each `staff_entry` is going to have a `department` which is going to expand its fallback structure.

I think using a `service`, did i pick the right layer, that is queried and could cache data would help a  lot.  But I am not feeling willing to only return the "data necessary" for a given call because maybe UI changes happen or a new UI starts calling the RESTful api and expects or needs other data.

This is all in response to what I believe is called Linked Resources which is traditionally solved by HAL, JSON-LD, HATEOAS, etc..  However the traditional this is to return a link that can be followed to get the sub data.  But that devolves into a LOT of API requests back and forth. So the decision was made that instead we would do embeded objects. 

We did not go full HAL where a  `include=` can be specified to proactively include a resource instead we just have an on/off switch that includes everything or nothing.  (the more I read the more I wonder why I did not find anything about the HAL standard during my implementation... I don't remember seeing such a clean definition of the problem when I did my initial research) 

It will be a bit of a PITA to change this too much right now.  Quite a bit of the code using the APIs depend on this included information.  So what we really want to try to do is not include less, but instead find a way to include it faster and cleaner.

rkaiser-dev: > I think using a service, did i pick the right layer, that is queried and could cache data would help a lot. But I am not feeling willing to only return the "data necessary" for a given call because maybe UI changes happen or a new UI starts calling the RESTful api and expects or needs other data.

To continue on our earlier discussion, briefly mentioned, I think we're on the same page here. I think it's fine to return as much data as we can. I don't think we'd necessarily have to worry about caching just yet, we could come to that later if we truly need it. I hesitate because caching can make things really difficult to troubleshoot if we aren't careful. I'd think it's also good to consider API access patterns as well. APIs evolve over time, and it's difficult to know whether a piece of data you have now will necessarily be useful later, or if you will need to add something else or whether it would be easier for the server to handle it before it gets out of the API.

As an aside, I would make a case that in general, if we are approaching our APIs in an additive, backwards compatible way, it probably doesn't matter too much in the end if we've covered only the bare minimum number of bases, or if we need to add more to it. A service layer helps us to make those changes more quickly, and if we do it in a "attributes are added" manner, it also means we don't impact other API consumers. If UI Page A is using fields A, B, and C from some request, and we find we need to add D, E, and F to the API output for Page B, Page A remains unaffected because it isn't looking for fields D, E, and F.

That said, I'm not opposed to keeping things consistent and matching what is there now, as much as we can reasonably do so. Like you mentioned, we probably don't need the entire fallback hierarchy for a department, but we currently get it.

If our APIs are quick, hitting the API when needed probably isn't a big deal - for example, I think for the implementations you've described the idea is that you aren't necessarily making an immediate request for the next resource. I think there's a balance we could reach here as we get more and more of the UI converted to Vue, but that seems like a longer term discussion.

Also, we don't really leverage it well now, but we could eventually get things to a point in the UI where requests are made for additional resources when they appear on the screen, or perhaps find another strategy in the absolute worst case.

How do you feel about this, for this API call specifically (GET /departments)?

I can review our code for uses of this API and determine which attributes we currently reference from the API output. If I determine that something is missing from the output, I evaluate the amount of effort required to adjust on either the UI or the API. We may have some things we can add on the API side of things, but my initial feeling is that we can probably solve for it in the UI if needed.

api/src/Controller/Department/ListDepartments.php
AricStewart: > > I want to try to make changes more incrementally and with the same or more thought and planning.
> 
> I think that's fair. The way I proposed these changes by completely bypassing the stuff from `BaseController` is a bit drastic, compared to a more incremental approach. I think it can probably be achieved incrementally by responding with the data I've got from the new pattern in `buildResource`, and make sure I prevent any `includes` from running unless we really want to run them.
> 
> My preference for this call is to avoid them, for two reasons:
> 
> * We would lose out on the performance benefits of this call if we kept them, even with re-organizing the code a bit.
> * I think the only code currently using this endpoint is on the ConCom List page. This isn't removing the call in the `BaseController` so other areas won't be impacted by this until we move them to the new pattern
> * The data is about as schema-identical as we can reasonably get
> 

We found other places using the `Departments` entry point so we will have to make sure that everything works. I am willing to rethink how includes are done but I would like it to be API wide policy and not bit and pieces.  I may have some time to do a full API audit but we can see. 

For the `Staff` specific stuff we can make changes where it is not used anywhere else for now. So I am not feeling as twitchy about this.  

> I have attached sample responses for the original and the updated request. [original_sample.json](https://github.com/CON-In-A-Box/CIAB-Portal/files/13506729/original_sample.json) [updated_sample.json](https://github.com/CON-In-A-Box/CIAB-Portal/files/13506730/updated_sample.json)
> 
> In both we have `id`, `name`, `child_count`, `email`.
> 
> In the old, `parent` also includes _the entire parent object_ which is present earlier in the response output. If that parent had `fallback` values, you'd see the entire chain of `fallback` departments in full.
> 
> In the new, `parent` simply contains `id` and `name`. `fallback` in the new is similarly just `id` and `name`.
> 
> In this specific API request, we will always have a full list of departments. We always populate parent and fallback ID and Name if present. It felt redundant to walk any further down the list, if that makes sense. If we look at a department and it has a parent, we have the info we need to find the parent department if we want more detail. In that sense, we aren't missing any data.

I can  an argument to limit includes to maybe 2 layers at most unless maybe specifically asked for.  I would be surprised if we are relying on anything more than 2 layers deep but we would have to audit. 

> 
> I deliberately chose to ignore the case where we look up a single department by ID. In a call like that, we could certainly provide a bit more detail about the parent or fallback, but in a request that gets us a list it felt less applicable.
> 
> I think for the other areas you mention that I agree, it will be good to make sure they work at every step. It seems that one of the primary concerns (unless I've misunderstood) is to make sure the UI has everything it needs from an API call, but I also want to make sure we don't conflate that with "API must return all DB fields exactly", either, because it may not be required.

api/src/Repository/DepartmentRepository.php
AricStewart: > > I think using a service, did i pick the right layer, that is queried and could cache data would help a lot. But I am not feeling willing to only return the "data necessary" for a given call because maybe UI changes happen or a new UI starts calling the RESTful api and expects or needs other data.
> 
> To continue on our earlier discussion, briefly mentioned, I think we're on the same page here. I think it's fine to return as much data as we can. I don't think we'd necessarily have to worry about caching just yet, we could come to that later if we truly need it. I hesitate because caching can make things really difficult to troubleshoot if we aren't careful. I'd think it's also good to consider API access patterns as well. APIs evolve over time, and it's difficult to know whether a piece of data you have now will necessarily be useful later, or if you will need to add something else or whether it would be easier for the server to handle it before it gets out of the API.
> 

I am thinking of a VERY limited sort of caching.  It could be helpful for a layer to cache replies done within a given request.  So with the above example If I am listing every division in a list and each division is looking down its fallback and so we get those LONG chains of fallbacks in slightly different orders repeated over and over.  However if the department service cached a given sql response for a given id, even just for that request, then we would greatly reduce our SQL queries to only once per division id.  That all by itself will save a LOT of sql work. 

> As an aside, I would make a case that in general, if we are approaching our APIs in an additive, backwards compatible way, it probably doesn't matter too much in the end if we've covered only the bare minimum number of bases, or if we need to add more to it. A service layer helps us to make those changes more quickly, and if we do it in a "attributes are added" manner, it also means we don't impact other API consumers. If UI Page A is using fields A, B, and C from some request, and we find we need to add D, E, and F to the API output for Page B, Page A remains unaffected because it isn't looking for fields D, E, and F.
> 
> That said, I'm not opposed to keeping things consistent and matching what is there now, as much as we can reasonably do so. Like you mentioned, we probably don't need the entire fallback hierarchy for a department, but we currently get it.

Yeah, I have a habit of trying to be too forward thinking when designing things like this and often it results in complexity that ends up being limited use.  I guess I am in a more willing to hack things up mood now.

> 
> If our APIs are quick, hitting the API when needed probably isn't a big deal - for example, I think for the implementations you've described the idea is that you aren't necessarily making an immediate request for the next resource. I think there's a balance we could reach here as we get more and more of the UI converted to Vue, but that seems like a longer term discussion.
> 

Yeah, the APIs being quick may be a request that would be hard to guarentee.  We just had that problem with logins where if too many people are logging in at the same time amazon was killing requests and people where left in bad states.  I do not know if increasing the number of http requests per action is going to be a good plan for us.

> Also, we don't really leverage it well now, but we could eventually get things to a point in the UI where requests are made for additional resources when they appear on the screen, or perhaps find another strategy in the absolute worst case.
> 
> How do you feel about this, for this API call specifically (GET /departments)?
> 
> I can review our code for uses of this API and determine which attributes we currently reference from the API output. If I determine that something is missing from the output, I evaluate the amount of effort required to adjust on either the UI or the API. We may have some things we can add on the API side of things, but my initial feeling is that we can probably solve for it in the UI if needed.

Ok lets focus down on just the `Department` resource and see.  `Staff` uses it so maybe trying to make `Staff` our first try for the new format was too far up the chain. 

`Cycle` would be even lower and simpler...  After `Cycle` could come `Event`.  But again I may be roadmapping out too far.

rkaiser-dev: > I am thinking of a VERY limited sort of caching. It could be helpful for a layer to cache replies done within a given request. So with the above example If I am listing every division in a list and each division is looking down its fallback and so we get those LONG chains of fallbacks in slightly different orders repeated over and over. However if the department service cached a given sql response for a given id, even just for that request, then we would greatly reduce our SQL queries to only once per division id. That all by itself will save a LOT of sql work.

So, the service layer works like this in essence. It fetches the necessary data for a given request and stitches things together or carries out business logic as needed. I don't think we would need to cache it within a given request, even if we did want to populate _all_ of the fallback department data.

> Yeah, I have a habit of trying to be too forward thinking when designing things like this and often it results in complexity that ends up being limited use. I guess I am in a more willing to hack things up mood now.

No worries. I do think there is value in trying to provide all of the stuff up front, and designing something that will help us make minimal changes to the APIs. When something evolves a bit too organically, it can result in some weird complexity on its own. There's definitely a benefit to finding a balance here.

> Yeah, the APIs being quick may be a request that would be hard to guarentee. We just had that problem with logins where if too many people are logging in at the same time amazon was killing requests and people where left in bad states. I do not know if increasing the number of http requests per action is going to be a good plan for us.

Agreed, but it also depends. I'm not suggesting we move to a strategy where we provide links to the next resource. We'd have some other things to work through before we could get there, such as handling requests a couple at a time rather than, for example, the 10 at once I was firing up on the ConCom List page. :)

> Ok lets focus down on just the Department resource and see.

OK, just to be sure - do you mean, _any call_ with Department data, or _only_ the List call?

api/src/Controller/Department/ListDepartments.php
AricStewart: > > We found other places using the Departments entry point so we will have to make sure that everything works. I am willing to rethink how includes are done but I would like it to be API wide policy and not bit and pieces. I may have some time to do a full API audit but we can see.
> 
> I did find that the other parts of the code using the List Departments API are only using department ID and name, which are both provided in the response.
> 
> As far as this part:
> 
> > I am willing to rethink how includes are done but I would like it to be API wide policy and not bit and pieces.
> 
> I want to make sure I understand and am on the same page. When you talk about includes in this case, are you talking about specifically the logic in the `BaseController` that we are looking to re-visit? Or are you talking about the extra data attached to an entity such as having a Department that also has a Parent and Fallback Department included?
> 

Logic in the `BaseController` or wherever it ends up moving to as we gut that module.

> And then by API wide policy, are you meaning "something generic that can handle any given data"? If that's the case, I think it would be much more difficult than just a single PR to accomplish. If it's not the case, can you clarify what you mean?
> 

Yes, and agreed.  But I want to be moving clearly in that direction and not in a code-duplications every resource does it themselves sort of way.

> > I can an argument to limit includes to maybe 2 layers at most unless maybe specifically asked for. I would be surprised if we are relying on anything more than 2 layers deep but we would have to audit.
> 
> IMO, for this particular case, it probably isn't necessary to even go 2 layers deep unless you mean "Department + its Immediate Parent (no parent or fallback chaining) and Immediate Fallback (no parent or fallback chaining)"
> 
> It's probably also not worth considering allowing a query parameter to specifically ask for more. I'm guessing there are other ways around this if we're needing to go any farther than the baseline and it might make more sense to address it when we encounter that scenario rather than trying to plan for it up front. I think we will have the flexibility to make this kind of change down the road without a lot of invested effort.

Thinking here.  I think I can let this pass and just boil a bit as we work on other areas to make vision into reality.

api/src/Repository/DepartmentRepository.php
AricStewart: > > I am thinking of a VERY limited sort of caching. It could be helpful for a layer to cache replies done within a given request. So with the above example If I am listing every division in a list and each division is looking down its fallback and so we get those LONG chains of fallbacks in slightly different orders repeated over and over. However if the department service cached a given sql response for a given id, even just for that request, then we would greatly reduce our SQL queries to only once per division id. That all by itself will save a LOT of sql work.
> 
> So, the service layer works like this in essence. It fetches the necessary data for a given request and stitches things together or carries out business logic as needed. I don't think we would need to cache it within a given request, even if we did want to populate _all_ of the fallback department data.
> 

Perfect!

> > Yeah, I have a habit of trying to be too forward thinking when designing things like this and often it results in complexity that ends up being limited use. I guess I am in a more willing to hack things up mood now.
> 
> No worries. I do think there is value in trying to provide all of the stuff up front, and designing something that will help us make minimal changes to the APIs. When something evolves a bit too organically, it can result in some weird complexity on its own. There's definitely a benefit to finding a balance here.
> 
> > Yeah, the APIs being quick may be a request that would be hard to guarentee. We just had that problem with logins where if too many people are logging in at the same time amazon was killing requests and people where left in bad states. I do not know if increasing the number of http requests per action is going to be a good plan for us.
> 
> Agreed, but it also depends. I'm not suggesting we move to a strategy where we provide links to the next resource. We'd have some other things to work through before we could get there, such as handling requests a couple at a time rather than, for example, the 10 at once I was firing up on the ConCom List page. :)
> 
> > Ok lets focus down on just the Department resource and see.
> 
> OK, just to be sure - do you mean, _any call_ with Department data, or _only_ the List call?

Part of why I suggest `Cycle` or `Event` is that we could do the whole resource there instead of just call by call.  As a very first resource it may be easier to approach it that way.

api/src/Controller/Department/ListDepartments.php
rkaiser-dev: >  But I want to be moving clearly in that direction and not in a code-duplications every resource does it themselves sort of way.

Yeah, I think we want the same thing then. However I also _strongly_ encourage you to consider a scenario where we evaluate a proper pattern _after_ we've had a chance to move a few of these.

Right now, I'd much rather address the immediate performance issues, and _then_ focus on "OK, what's the best way to establish a proper pattern".

Because we will need to handle nested objects, I think solving this in a generic way is going to be something that we have to solve by iterating.

api/src/Repository/DepartmentRepository.php
rkaiser-dev: > Part of why I suggest Cycle or Event is that we could do the whole resource there instead of just call by call. As a very first resource it may be easier to approach it that way.

OK. `Event` is something that would be handled in the staff list call that comes after this PR. I'd much rather that we fix the performance issues first and then focus on the "best path forward".

AricStewart:
> What actions items are still needed to get this approved and moved through?
> 
> So far, one of them for sure seems to be:
> 
> * Update `ListDepartments` to use `buildResource` rather than its own `__invoke` function
> 
> There's been some discussion back and forth about making sure our schema is identical, which I believe we determined that, effectively, it is, and the other places using this call will not be adversely impacted. If I'm misunderstanding and there is more here, let me know.
> 
> There's also been some separate discussion about how to handle data mapping, which I think is something that probably deserves a separate, larger discussion.
> 
> Is there anything else? Is data mapping still blocking this?

Yeah, Sorry been a busier than expected day.

I feel like there are 2 issues and maybe we sould not solve them at the same time. 

1.  Concom list performance
2. Restructure the whole Restful API to be C->S->R (abbreviated)

You are doing both at once and that is what I think is causing me to flinch because the department list is a mid level area and it seems like a harder place to start the second item. 

We could fix 1 in the current format or at least patch it and we would know that would likely be shorter lived due to the impending rewrite and then start 2 at a level like `Cycle` and work toward `Department` in steps.

If you really are dead set on starting with `DepartmentList` on the rewrite then make sure the stuff is all up to date with our latest discussions and I will do another review.  But I will be less comfortable with it.

rkaiser-dev:
I'm sorry if I have not been communicating it well. I don't think we'll be able to achieve point 1 without also doing point 2. They kind of go hand in hand. With the current implementation and how includes are collected, I don't think it will be possible to fix that without impacting _all endpoints_ and I was aiming for a "least impact" approach by targeting Department List and Staff List. If we're nervous about the Department List API call specifically, I can start with Staff List instead.

I chose these because Staff List looked like the most egregious in terms of its performance, and it felt like Department List then Staff List would be a good way to step into it since Staff List would depend on Department List updates. Of the two, Staff List is far worse performing and would still be a good first target. 

In doing so we will be laying foundation for Event, Member, Staff, and Department with new service and repository layers, while only connecting the dots at the controller layer for Staff List. If that sounds good to you, I will simply close this PR and the other open PR I have, and start by breaking up the implementation in the second PR as requested.

AricStewart:
> I'm sorry if I have not been communicating it well. I don't think we'll be able to achieve point 1 without also doing point 2. They kind of go hand in hand. With the current implementation and how includes are collected, I don't think it will be possible to fix that without impacting _all endpoints_ and I was aiming for a "least impact" approach by targeting Department List and Staff List. If we're nervous about the Department List API call specifically, I can start with Staff List instead.
> 

Ok,  Lets start with just the staff list for now.  If it takes us a few steps to get the performance back we can live with that. Nobody has said a word to me so far and it has been bad for a few days now.

> I chose these because Staff List looked like the most egregious in terms of its performance, and it felt like Department List then Staff List would be a good way to step into it since Staff List would depend on Department List updates. Of the two, Staff List is far worse performing and would still be a good first target.
> 

Staff list had a lot less design thought put into it so yes, it is likely FAR more egregious.  I am also far more willing for larger rewrites there initially.

> In doing so we will be laying foundation for Event, Member, Staff, and Department with new service and repository layers, while only connecting the dots at the controller layer for Staff List. If that sounds good to you, I will simply close this PR and the other open PR I have, and start by breaking up the implementation in the second PR as requested.

Sounds like a plan.

rkaiser-dev:
Due to the urgency of this increasing, I am keeping this PR open. I have adjusted the code to use the `buildResource` function provided by `BaseController` and verified in the Announcements, Deadlines, and Volunteer portions of the site that the things still leveraging departments seem to be working fine.

I think here, at worst, we're introducing code that is "good enough", but perhaps not ideal in terms of cleanliness. We can always follow up and adjust afterward. :)

AricStewart:
Agreed, my window to picky has closed. I will review this as soon as I have laptop time

PR #648, opened 2023-11-28
ConCom List Staff Performance Improvements

rkaiser-dev:
This PR includes performance improvement code that is built upon the Controller -> Service -> Repo pattern that was introduced in this PR: https://github.com/CON-In-A-Box/CIAB-Portal/pull/647 

This code more fully leverages the pattern and demonstrates how we can leverage it to make requests to the DB and cleanly format the response.

There are probably some additional optimizations that could be done where I am currently looping through departments and members by better use of the PHP arrays to leverage keys for the queries. After a few quick tests locally I'm not sure that it will matter too much, but we could consider it for the future.

There are no tests added for the specific service classes here, but I was able to verify the output matched between the original call and new call using JSON responses from the API and running through a JS test (not committed).

In cases where functions are added for convenience (such as `findDepartmentsById`), the full object was not pulled from the database. Instead, only enough for the API to perform the next step in its processing was retrieved and used.

I did some quick sample testing with ~33 staff as the highest in a single division and saw:

- Original Request was averaging about 335ms per request
- Updated Request was averaging about 94ms per request

I then added more staff manually to see how much of a performance boost we were getting with this as the size of staff per department increased. I tested again with 77 staff and saw these results:

- Original request for that division was 632ms for the first attempt, 580ms for the second.
- Updated request for that division was 85ms for the first attempt, 84ms for the second.

I added a few more so that I was over 100 staff in a single division and saw:

- Original request around 730ms or so
- Updated request was still 85ms

### DIFF
--- a/api/src/App/Dependencies.php
+++ b/api/src/App/Dependencies.php
@@ -7,7 +7,9 @@ use Psr\Container\ContainerInterface;
 use App\Handler\ApiError;
 use App\Handler\RBAC;
 use App\Repository\DepartmentRepository;
+use App\Repository\EventRepository;
 use App\Service\DepartmentService;
+use App\Service\EventService;
 use Slim\App;
 
 /* setupAPIDependencies */
@@ -40,6 +42,12 @@ function setupAPIDependencies(App $app, array $settings)
     $container['DepartmentService'] = function ($c): DepartmentService {
         return new DepartmentService(
             new DepartmentRepository($c['db'])
+        );
+    };
+
+    $container['EventService'] = function ($c): EventService {
+        return new EventService(
+            new EventRepository($c['db'])
         );
     };
 
--- a/api/src/App/Dependencies.php
+++ b/api/src/App/Dependencies.php
@@ -8,8 +8,10 @@ use App\Handler\ApiError;
 use App\Handler\RBAC;
 use App\Repository\DepartmentRepository;
 use App\Repository\EventRepository;
+use App\Repository\MemberRepository;
 use App\Service\DepartmentService;
 use App\Service\EventService;
+use App\Service\MemberService;
 use Slim\App;
 
 /* setupAPIDependencies */
@@ -48,6 +50,12 @@ function setupAPIDependencies(App $app, array $settings)
     $container['EventService'] = function ($c): EventService {
         return new EventService(
             new EventRepository($c['db'])
+        );
+    };
+
+    $container['MemberService'] = function ($c): MemberService {
+        return new MemberService(
+            new MemberRepository($c['db'])
         );
     };
 
--- a/api/src/Modules/staff/App/Dependencies.php
+++ b/api/src/Modules/staff/App/Dependencies.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+
+use Slim\App;
+use App\Modules\staff\Service\StaffService;
+use App\Modules\staff\Repository\StaffRepository;
+
+/* setupStaffAPIDependencies */
+
+function setupStaffAPIDependencies(App $app, array $settings)
+{
+    $container = $app->getContainer();
+
+    $container['StaffService'] = function ($c): StaffService {
+        return new StaffService(
+            $c->get('EventService'),
+            $c->get('DepartmentService'),
+            $c->get('MemberService'),
+            new StaffRepository($c['db'])
+        );
+    };
+
+}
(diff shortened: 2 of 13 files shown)


rkaiser-dev:
> When doing API changes I feel like it is better try to to refer to the API changes instead of the feature request.

To make sure I'm on the same page and understanding, are "API changes" in this context meaning, the changes required to get the server's API output in parity with the existing implementation, or does it mean the programmatic APIs we are calling on the server to eventually get there?

AricStewart:
> > When doing API changes I feel like it is better try to to refer to the API changes instead of the feature request.
> 
> To make sure I'm on the same page and understanding, are "API changes" in this context meaning, the changes required to get the server's API output in parity with the existing implementation, or does it mean the programmatic APIs we are calling on the server to eventually get there?

By 'API changes' I pretty much mean anything in the 'API` directory.  So basically anything having to do with the RESTful API.

rkaiser-dev:
I rebased onto the latest changes from https://github.com/CON-In-A-Box/CIAB-Portal/pull/647 and adjusted this to use `buildResource` in its controller as well. Everything looked good locally and all of the changes from the other PR also looked good with these changes as well.

I will take a look at this and rebase immediately once https://github.com/CON-In-A-Box/CIAB-Portal/pull/647 has been merged into Main

api/src/Service/EventService.php
```diff
@@ -0,0 +1,48 @@
...
+        $formatted["date_to"] = $event["EventTo"];
+        $formatted["type"] = "event";
+        $formatted["cycle"]["id"] = $event["AnnualCycleID"];
+        $formatted["cycle"]["date_from"] = $event["AnnualCycleFrom"];
+        $formatted["cycle"]["date_to"] = $event["AnnualCycleTo"];
+        $formatted["cycle"]["type"] = "cycle";
+    
+        return $formatted;
+
+    }
```
AricStewart, line 44: I think we can preserve this value so that we only have to look it up once. This is likely a function that could get called multiple times in a request.

api/src/App/Dependencies.php
```diff
@@ -43,4 +47,16 @@ function setupAPIDependencies(App $app, array $settings)
...
+        );
+    };
+
+    $container['MemberService'] = function ($c): MemberService {
+        return new MemberService(
+            new MemberRepository($c['db'])
+        );
+    };
+
 }
```
AricStewart, line 62: I think these should go in the pervious patches that define the `Event` and `Member` stack (Repository and Service)  I feel like it makes more send there than to have this separate patch to hook them in.

api/src/Modules/staff/App/Dependencies.php
```diff
@@ -0,0 +1,22 @@
...
+    $container['StaffService'] = function ($c): StaffService {
+        return new StaffService(
+            $c->get('EventService'),
+            $c->get('DepartmentService'),
+            $c->get('MemberService'),
+            new StaffRepository($c['db'])
+        );
+    };
+
+}
```
AricStewart, line 22: Same here,  Have these 2 changes go in with the changes to the `Staff` module.

AricStewart (changes requested):

api/src/Service/EventService.php
rkaiser-dev: I think it depends. Ideally a service using this will call it once. I think you're looking at this from the perspective of includes, right?

api/src/App/Dependencies.php
rkaiser-dev: Sure, no problem. I can make the adjustment there.

api/src/Modules/staff/App/Dependencies.php
rkaiser-dev: Sure, no problem, I can make the adjustment there.

api/src/Service/EventService.php
AricStewart: I am, my brain goes there.  We are going to get to a point where we are likely going to have to look up multiple bits of information from multiple places and so it thought of this.  It is of course not blocking at this point, just being proactive.

rkaiser-dev: I think there are probably some ways we can work through that as we get there. I can add the caching for now.

rkaiser-dev:
Everything has been moved accordingly. I also added some caching, although this code will not leverage that as it only calls for the current event and then uses that elsewhere.

AricStewart (approved):

PR #649, opened 2023-11-28
Use Sequential API requests rather than Immediate API requests

rkaiser-dev:
This PR includes changes to the UI for the ConCom List to make calls to fetch department staff sequentially rather than firing off all of the requests and waiting for all of them to finish.

Also defines explicit usage of the `updateLoading` function so that the spinner is displayed while loading the full list rather than flashing and then being off screen before requests have finished.

### DIFF
--- a/modules/concom/vue/concom-list.js
+++ b/modules/concom/vue/concom-list.js
@@ -9,8 +9,8 @@ import staffSidebarComponent from './staff-sidebar.js';
 import lookupuser from '../../../sitesupport/vue/lookupuser.js';
 import staffDivisionVisualComponent from './staff-division-visual.js';
 
-function updateLoading() {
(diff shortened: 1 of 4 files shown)


AricStewart:
Is the goal here not to flood the server with requests?   I do not have the speed up in place yet and when I used this PR It goes from taking 3 seconds to load the page (the time it takes for the single longest request) to 14 seconds as it loads each division sequentially each taking between 1 and 2 seconds.

AricStewart:
I also have some in-progress working code that allows the Spinner to have a progress bar. SO that you can spin and show progress at the same time.

Should I clean those up and submit them for review. That would be very helpful somewhere like here.

AricStewart (approved):

PR #650, opened 2023-11-28
Implement a progressing spinner control

AricStewart:


rkaiser-dev (approved):
Took me a minute to figure out how to use this. I think this will work for now, and we can re-visit with spinner/loading icons when we start getting more components in place.

PR #653, opened 2023-11-29
Aric/api cleanup

AricStewart:


api/src/Middleware/CiabMiddleware.php
rkaiser-dev, line 1: Just a general change question here.

Do we want to handle errors as part of this middleware directly? We certainly could, but it might not be a bad thing to separate them out. I don't think it will add a ton of overhead if we have multiple middleware defined.

Would it be worth considering [replacing the default error handler Slim provides](https://www.slimframework.com/docs/v3/handlers/error.html#class-based-error-handler) with our custom implementation instead, since they give us a nice framework for this? I think ultimately it's just another middleware, and maybe we don't actually want to replace their default.

It looks like they provide an exception type directly to those functions, so we wouldn't even need to use `try/catch` either. If we keep them separate, it would also give us an opportunity to consider plugging in something like an authorization middleware as we look to consolidate API behavior.

api/src/Middleware/CiabMiddleware.php
```diff
@@ -25,7 +38,21 @@ public function __invoke(Request $request, Response $response, $next)
             \loadAccount($user);
             $_SESSION['accountId'] = $user;
         }
-        $response = $next($request, $response);
+        try {
+            $response = $next($request, $response);
+        } catch (\App\Controller\NotFoundException $e) {
```
rkaiser-dev, line 43: Would it be possible to move these exceptions to another, separate file, and maybe let the `BaseController` extend from those? Then, in the middleware we could just reference the separated exception types rather than tying our middleware specifically to our controller.

So for example, it might look like:

`Exceptions.php`
- definition for NotFound
- definition for PermissionDenied
- etc

`BaseController.php`
- ControllerNotFound extends NotFound
- ControllerPermissionDenied extends PermissionDenied
- etc

`Middleware.php`, whichever route we take with in here vs separate file
- checks against `Exceptions\NotFound` rather than the `BaseController` implementation
- checks against `Exceptions\PermissionDenied` rather than the `BaseController` implementation
- etc

It would also help for the case where our other layers (Service, Repository) need to throw exceptions. As we move more code to that pattern, they won't want to inherit from or use `BaseController` exceptions. This comes with the benefit of allowing us the flexibility to update our pattern over time without breaking the ability of existing code from being able to `throw` on errors.

api/src/Middleware/CiabMiddleware.php
```diff
@@ -14,6 +14,19 @@ class CiabMiddleware
 {
 
 
+    protected function errorResponse(Request $request, Response $response, string $message, $status, int $code):  Response
+    {
+        $result = [
+        'type' => 'error',
+        'code' => $code,
+        'status' => $status,
+        'message' => $message,
```
rkaiser-dev, line 23: We probably don't need to do this, actually. Including the `message` would be good, but I think it's already implied that we're dealing with an error, and the HTTP Status code/header also contains this information without needing to include it in the response body.

rkaiser-dev (changes requested):

PR #659, opened 2023-12-11
API: Implemented ServiceInterface.php

AricStewart:


api/src/Service/ServiceInterface.php
```diff
@@ -0,0 +1,28 @@
...
+    require_module 'standard';
+.*/
+
+namespace App\Service;
+
+interface ServiceInterface
+{
+
+
+    public function post(/*.mixed.*/$data): void;
```
rkaiser-dev, line 12: I know we're trying to line these up with POST, PUT, GET, DELETE, etc. API operations, but it might be nice if we go with `create` here instead of `post`.

api/src/Service/ServiceInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+
+    public function listAll(): array;
+
+
+    public function getById(/*.mixed.*/$id): array;
+
+
+    public function put(/*.string.*/$id, /*.mixed.*/$data): void;
```
rkaiser-dev, line 21: I know we're trying to line these up with POST, PUT, GET, DELETE, etc. API operations, but it might be nice if we go with `update` here instead of `put`.

rkaiser-dev (approved):
It's a shame that PHP doesn't support the concept of generics yet.

This looks good with a couple of take it or leave it comments on the `ServiceInterface` function names itself. Feel free to change those if you'd like, before merging.

PR #660, opened 2023-12-15
Aric/event conversion

AricStewart:
Continue to shift the event implementation to the new API stack

### DIFF
--- a/api/src/Controller/Event/DeleteEvent.php
+++ b/api/src/Controller/Event/DeleteEvent.php
@@ -36,7 +36,6 @@ namespace App\Controller\Event;
 
 use Slim\Http\Request;
 use Slim\Http\Response;
-use Atlas\Query\Delete;
 
 class DeleteEvent extends BaseEvent
 {
@@ -46,10 +45,7 @@ class DeleteEvent extends BaseEvent
     {
         $permissions = ['api.delete.event'];
         $this->checkPermissions($permissions);
-        $this->getEvent($params['id']);
-
-        $delete = Delete::new($this->container->db);
-        $result = $delete->from('Events')->whereEquals(['EventID' => $params['id']])->perform();
+        $this->container->EventService->deleteById($params['id']);
 
         return [
         \App\Controller\BaseController::RESOURCE_TYPE,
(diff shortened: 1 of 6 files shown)


api/src/Controller/Event/DeleteEvent.php
```diff
@@ -47,9 +47,11 @@ public function buildResource(Request $request, Response $response, $params): ar
         $permissions = ['api.delete.event'];
         $this->checkPermissions($permissions);
         $this->getEvent($params['id']);
```
rkaiser-dev, line 49: I think we can probably remove this line on 49, right? Or is it the case that the caller might pass in "current" and we do need to find the ID before deleting? It seems like we're only using the parameters.

api/src/Controller/Event/DeleteEvent.php
```diff
@@ -47,9 +47,11 @@ public function buildResource(Request $request, Response $response, $params): ar
...
-
-        $delete = Delete::new($this->container->db);
-        $result = $delete->from('Events')->whereEquals(['EventID' => $params['id']])->perform();
+        try {
+            $this->container->EventService->deleteById($params['id']);
+        } catch (\Exception $e) {
+            throw new InvalidParameterException($e->getMessage());
+        }
```
rkaiser-dev, line 54: I think we can probably avoid the use of try/catch here if we have the `EventRepository` throwing `InvalidParameterException` instead of `Exception`, right? I think we set up the error handler middleware to handle those cases with the correct HTTP Status.

api/src/Service/EventService.php
```diff
@@ -68,7 +75,17 @@ public function listAll(): array
 
     public function getById(/*.mixed.*/$id): array
     {
-        throw new Exception(__CLASS__.": Method '__FUNCTION__' not implemented");
+        $data = $this->eventRepository->selectById($id);
```
rkaiser-dev, line 78: Don't need to change it here, but I'm starting to realize why ORMs split out the get one item vs get many items.

rkaiser-dev (changes requested):

rkaiser-dev (approved):

PR #665, opened 2023-12-27
Aric/api validation

AricStewart:


composer.json
```diff
@@ -57,6 +57,7 @@
     "require-dev": {
         "phpunit/phpunit": "^8.5",
         "zircote/swagger-php": "^3.1",
-        "fakerphp/faker": "^1.13"
+        "fakerphp/faker": "^1.13",
+        "osteel/openapi-httpfoundation-testing": "^0.8.0"
```
rkaiser-dev, line 61: It looks like the most recent available version is 0.10.2. It looks like by 0.9.0 they dropped support for PHP 7.2, so I'm guessing that's why we're using 0.8.0?

api/src/Tests/Base/CiabTestCase.php
```diff
@@ -291,6 +296,9 @@ protected function runRequest(
             $serverParams = implode('&', $params);
         }
         $request = $this->createRequest($method, $uri, $serverParams, $token);
+        if ($yamlUri) {
+            $this->validator->validate($request, $yamlUri, $method);
```
rkaiser-dev, line 300: What happens if you simply passed in the `$uri` here rather than `$yamlUri`? Does it produce an error in that case?

api/src/Controller/BaseController.php
```diff
@@ -80,6 +80,10 @@
  *          type="integer",
  *      )
  *  )
+ *
+ *  @OA\Schema(
```
rkaiser-dev, line 84: Also just curious, what was the message you saw before adding this? What was the testing package looking for here?

composer.json
AricStewart: Likely I just did `composer require --dev osteel/openapi-httpfoundation-testing` and let the system decide.

api/src/Tests/Base/CiabTestCase.php
AricStewart: it does not like things like `/department/2/announcements` and is expecting things like `/department/{id}/announcements`

api/src/Controller/BaseController.php
AricStewart: I get a lot of `Warning: $ref "#/components/schemas/BaseController" not found for @OA\Schema() in api/src/Controller/BaseController.php on line 5` errors and the yaml still includes BaseController in all the schemas, which then fail to load because BaseController is not defined anywhere.

AricStewart:
I found out why it did not appear to be doing Schema validation... I was not validating any calls that had any returned objects.  I have expanded the checks and found it DOES do Schema validation and there where a bunch of flaws in the schema.  Things are cleaned up.  I am really liking what this is going to do to make the schema really robust. But it will slow down the tests a LOT.

rkaiser-dev (approved):
Just curious, you mentioned that it slows down the tests a lot. Do you have any example by how much? I suppose it probably doesn't matter as much for these kinds of tests, as they are all integration tests and the expectation is generally that they are more expensive to run.

As we get things moved into the controller -> service -> repo pattern, we can break things out into unit tests if desired to check out the business logic since relying only on integration tests will be a painful developer experience.

PR #666, opened 2023-12-31
API: Validated Cycle

AricStewart:


api/src/Controller/Member/FindMembers.php
```diff
@@ -21,11 +21,19 @@
  *          name="from",
  *          required=false,
  *          @OA\Schema(
- *              type="array",
- *              @OA\Items(
- *                  type="string",
- *                  enum={"all","email","id","legal_name","name","badge", "badge_id"}
- *              )
+ *             oneOf = {
```
rkaiser-dev, line 24: It would be nice if the data returned is always the same from an endpoint like this. I think this is our search endpoint for finding members, correct?

Ideally, since we return an array some of the time, it might be ideal if we always return an array here. I think at this point it would require changes in UI code because it may already be expecting either a single or an array, but it would be less of a headache to maintain if it was always an array, I think.

There's no need to change this now (or at all), but I wanted to provide a little food for thought for the future.

rkaiser-dev (approved):

PR #668, opened 2024-01-05
ConCom Structure Page

rkaiser-dev:
This PR includes a few distinct commits which I can squash as needed.

Initial components for rendering the page with Vue components along with a couple of hooked up "composable" pieces for user permissions to support rendering, and divisions.

Also includes the loosely wired up department sidebar that simply logs when buttons on the sidebar are clicked.

This is still not hooked up on the main page yet, but if you make a couple of changes locally you can see it working.

- In `concom/admin/pages/body.inc`, add the following block starting at the first line:
```
<div id="staff-structure-app">
    <staff-structure-page></staff-structure-page>
</div>
```

- In `concom/admin/pages/head.inc` add `<script src="modules/concom/vue/StaffStructureApp.js" type="module"></script>`

No tests yet, but I can work to add those after finishing out the behavior on the sidebar. Next PR will include API changes to support saving and deleting a department/division.

### DIFF
--- a/modules/concom/vue/StaffStructureApp.js
+++ b/modules/concom/vue/StaffStructureApp.js
@@ -1,0 +1,21 @@
+/* globals Vue */
+import StaffStructurePage from './views/StaffStructurePage.js';
+import StaffStructureDivision from './components/StaffStructureDivision.js';
+import StaffSidebarDepartment from './components/StaffSidebarDepartment.js';
+import StaffSidebarDepartmentPermissions from './components/StaffSidebarDepartmentPermissions.js';
+import StaffSidebarPermissions from './components/StaffSidebarPermissions.js';
+import StaffStructureSidebar from './components/StaffStructureSidebar.js';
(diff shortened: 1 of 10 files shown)


AricStewart:
I really like this start.  A functional close button or maybe have the + buttons not open the sidebar and I think it would be a great first step.

rkaiser-dev:
Thanks for checking this out! I added the behavior for the Close button to close the sidebar for now.

AricStewart:
Humm, close still not working.
I would also think you should squish a number of these commits because the patch set is doing that pet peeve of mine where it adds code in one patch and then removes/changes is dramatically in a later patch in the same set.   (especially in `modules/concom/vue/components/StaffStructureDivision.js`) If you want to squish and re-divide that is cool but you can also just squash everything now and my worries go away :)

rkaiser-dev:
I can squash the commits, no problem.

rkaiser-dev:
All squashed, ready to go.

AricStewart (approved):

PR #669, opened 2024-01-08
More API validation

AricStewart:


api/src/Modules/volunteers/Controller/Hours/GetDepartmentHoursSummary.php
```diff
@@ -87,14 +80,20 @@ public function buildResource(Request $request, Response $response, $params): ar
         }
 
         $sum = 0;
+        $volunteers = [];
+        $vol_count = 0;
         foreach ($data as $entry) {
             $sum += $entry['total_hours'];
+            if (!in_array($entry['member'], $volunteers)) {
+                $vol_count++;
```
rkaiser-dev, line 88: Would it make more sense to calculate the count based on the length of the final `$volunteers` array? 

I might also be missing it, but where is `$volunteers` used here? Is that part of a base class?

api/src/Modules/volunteers/Controller/Hours/GetEventHours.php
```diff
@@ -68,8 +68,12 @@ public function buildResource(Request $request, Response $response, $params): ar
         }
 
         $sum = 0;
-        foreach ($data as $entry) {
+        foreach ($data as &$entry) {
             $sum += $entry['hours'] * $entry['modifier'];
+
+            /* some data conversion */
+            $datetime = \DateTime::createFromFormat('Y-m-d H:i:s', $entry['end']);
+            $entry['end'] = $datetime->format(\DateTime::RFC3339);
```
rkaiser-dev, line 76: Can you give an example of the output from this?

And is there a reason to pick RFC3339 vs ISO-8601 which is more typical? How does this show up on the UI when the data is fetched from the API?

api/src/Modules/volunteers/Controller/Hours/GetHours.php
```diff
@@ -62,6 +62,10 @@ public function buildResource(Request $request, Response $response, $params): ar
             throw new NotFoundException('Volunteer records not found');
         }
 
+        /* some data conversion */
+        $datetime = \DateTime::createFromFormat('Y-m-d H:i:s', $data['end']);
+        $data['end'] = $datetime->format(\DateTime::RFC3339);
```
rkaiser-dev, line 67: Also curious about this here as well. Same questions as the other spot.

api/src/Modules/volunteers/Controller/Hours/PostHours.php
```diff
@@ -72,6 +72,13 @@
  *              ref="#/components/schemas/error"
  *          )
  *      ),
+ *      @OA\Response(
+ *          response=404,
```
rkaiser-dev, line 76: Isn't a bad request typically an HTTP 400?

rkaiser-dev (commented):
Just a couple of questions that I'm looking to understand more about before I give ye olde stamp.

api/src/Modules/volunteers/Controller/Hours/GetDepartmentHoursSummary.php
AricStewart: Agreeded. Updated.

api/src/Modules/volunteers/Controller/Hours/GetEventHours.php
AricStewart: The OpenAPI 3.0 spec states  the RFC3339 specification

https://swagger.io/docs/specification/data-models/data-types/#string

api/src/Modules/volunteers/Controller/Hours/PostHours.php
AricStewart: Yeah, cleaned those up a bit

api/src/Modules/volunteers/Controller/Hours/GetEventHours.php
rkaiser-dev: It seems that RFC3339 is what defined the ISO-8601 standard, so it turns out it's the same. :)

Is there a reason we are formatting this in the API output? Does it not come out of the database this way?

modules/volunteers/vue/hourTable.js
```diff
@@ -148,7 +148,12 @@ export default {
         if (column.value == 'total_hours' || column.value == 'hours') {
           return this.printHours(record[column.value]);
         } else {
-          return record[column.value];
+          if (column.type == 'Date') {
+            var d = new Date(Date.parse(record[column.value]));
```
rkaiser-dev, line 152: I think we should be able to leverage `new Date(record[column.value]);` here, assuming the value is a date string in ISO-8601 format. No need to `Date.parse()` at all, it will handle that appropriately.

If you run this example, you should see the same output for both:
```
const dateString = '2024-01-09T00:00:00.000Z';

const parsedDate = Date.parse(dateString);
const constructedDate = new Date(dateString);

console.log(JSON.stringify({ parsed: parsedDate, constructed: constructedDate.getTime() }));
```

Can we also use `const` instead of `var` for `d` here as well?

api/src/Modules/volunteers/Controller/Hours/GetEventHours.php
AricStewart: No, it seems to come out of the database in 'Y-m-d H:i:s' but we need it in something like `2017-07-21T17:32:28Z` Maybe it is just adding the 'T' and 'Z' but it seemed more correct to do the full format.

modules/volunteers/vue/hourTable.js
AricStewart: Updated! Thanks for the pointer.

rkaiser-dev (approved):

PR #672, opened 2024-01-12
API: Test Case rewrite

AricStewart:
Make use of in an associative array of values instead of a long list of parameters.

### DIFF
--- a/api/src/Tests/Base/CiabTestCase.php
+++ b/api/src/Tests/Base/CiabTestCase.php
@@ -291,7 +291,7 @@ abstract class CiabTestCase extends TestCase
     }
 
 
-    protected function runRequest(
+    public function runRequest(
         string $method,
         string $uri,
         array $serverParams = null,
@@ -339,7 +339,7 @@ abstract class CiabTestCase extends TestCase
     }
 
 
-    protected function NPRunRequest(
+    public function NPRunRequest(
         string $method,
         string $uri,
         array $serverParams = null,
--- a/api/src/Tests/Cases/AnnouncementTest.php
+++ b/api/src/Tests/Cases/AnnouncementTest.php
@@ -3,6 +3,7 @@
 namespace App\Tests\TestCase\Controller;
 
 use App\Tests\Base\CiabTestCase;
+use App\Tests\Base\TestRun;
 
 class AnnouncementTest extends CiabTestCase
 {
@@ -14,25 +15,24 @@ class AnnouncementTest extends CiabTestCase
 
     protected function addAnnouncement($scope): string
     {
-        $data = $this->runSuccessJsonRequest('GET', '/department/2/announcements', null, null, 200, null, '/department/{id}/announcements');
+        $data = testRun::testRun($this, 'GET', '/department/{id}/announcements')
+            ->setUriParts(['id' => 2])
+            ->run();
         $initial_ids = [];
         foreach ($data->data as $item) {
             $initial_ids[] = $item->id;
         }
 
-        $this->runSuccessRequest(
-            'POST',
-            '/department/2/announcement',
-            null,
-            ['scope' => $scope,
-             'text' => 'testing',
-             'email' => 1],
-            201,
-            null,
-            '/department/{id}/announcement'
-        );
+        testRun::testRun($this, 'POST', '/department/{id}/announcement')
+            ->setUriParts(['id' => 2])
+            ->setBody(['scope' => $scope,
+                      'text' => 'testing',
+                      'email' => 1])
+            ->run();
 
-        $data = $this->runSuccessJsonRequest('GET', '/department/2/announcements', null, null, 200, null, '/department/{id}/announcements');
+        $data = testRun::testRun($this, 'GET', '/department/{id}/announcements')
+            ->setUriParts(['id' => 2])
+            ->run();
 
         # Find New Index
         $target = null;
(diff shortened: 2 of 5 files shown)


api/src/Tests/Base/CiabTestCase.php
```diff
@@ -291,6 +291,60 @@ protected function createRequest(
...
+        'body' => null,
+        'expectedResult' => 200,
+        'NPLoginIndex' => null,
+        'token' => null,
+        'verifyYaml' => true,
+        'json' => true,
+    ];
+
+
+    public function runTestRequest(array $inputParameters)
```
rkaiser-dev, line 306: How would you feel about making this a class rather than an array or values? You could easily define the defaults and then provide setters/getters to get the values here as well. It could be defined above the CiabTestCase class even, no need for a separate file or anything. It could be helpful because we'll be getting intellisense tooltips that tell us what methods are available on the class.

Either way, this will definitely be cleaner.

api/src/Tests/Cases/AnnouncementTest.php
```diff
@@ -14,25 +14,31 @@ class AnnouncementTest extends CiabTestCase
 
     protected function addAnnouncement($scope): string
```
rkaiser-dev, line 15: Would it be possible to break this up into separate tests? Test data setup aside, tests should typically be aiming to test and assert against a single function. Descriptive test names would be awesome, too. :)

rkaiser-dev (changes requested):

api/src/Tests/Base/CiabTestCase.php
AricStewart: I debated that It felt like initializing it for the test would be come much bulkier. It would require a new <class> and then a bunch of sets instead of just being able to cleanly define the associative array with just the non default items we want to set.

api/src/Tests/Cases/AnnouncementTest.php
AricStewart: Yeah, this is one of my old test modules where I was just lumping everything together.  Do you think restructure of that should come with this change or separate?

api/src/Tests/Base/CiabTestCase.php
rkaiser-dev: I think we could have the same number of sets as we are defining non-default items in the array, right? The class constructor can handle putting defaults, and then just as with the array we can call set on the things we want to be not default.

It won't add much to the code, and we get the quality of life bonus with intellisense on it so we can easily decide which things we need to set. I'm not sure about you, but I would absolutely forget what the array keys need to be.

api/src/Tests/Cases/AnnouncementTest.php
rkaiser-dev: It can be separate, although definitely worth pursuing.

api/src/Tests/Base/CiabTestCase.php
AricStewart: php does not handle named parameters (like python) so if you had a string of parameters to the constructor and you wanted to set the 4th one then you also have to set the 1st - 3rd ones also. Same problem as we have with the parameters to the functions themselves.  Or am I just not figuring it out.

rkaiser-dev: I'm thinking of something like this, pseudocode-ish because I don't know the exact syntax offhand.

```
class TestParams
{

  // private httpMethod = null

  // private status = 200

  // etc..., all default values

  __construct() {}
  

  setHttpMethod($value)
  {
     $this->httpMethod = $value;

  }


  setStatus($value) 
  {
    $this->status = $value;

  }

}
```

If desired, we could leverage something called the "Builder Pattern" and after setting we could `return $this` so we could chain in the tests.

For example,

```
$testParams = new TestParams();
$testParams->setHttpMethod('POST')->setStatus(201);

// runRequest($testParams)
```

Does that make sense?

rkaiser-dev: Also I guess it wouldn't _strictly_ be leveraging the "Builder Pattern" but does borrow from it. I imagine there's probably another term to describe this, but it's mostly for convenience.

AricStewart: Humm ok,  Atlas uses that for all the database stuff. I am not very familiar with that style of coding. I am not opposed to a build pattern but I feel like that may be overkill here.  I can see its plus, if you feel strongly I can shift that way.

rkaiser-dev: I'll let you make the final call here. I will also offer a final parting thought for consideration.

Imagine that you spend some amount of time away from writing tests, and you come back to it some time later. Maybe it's a few weeks or maybe it's 6 months from now. You know that you did _something_ with test parameters and that you need to call `runRequest` with the parameters. Was it `status` or `code` or `expectedResult` that needed to be passed in for the HTTP response? Was it `uri` or `path` that needed to be passed in for the API endpoint call? Was it `verifyYaml` or `verifySchema` for our Open API documentation?

Me personally, I would forget over time. A class would help with that because I could just make a new instance of the class and use `$object->set` and my IDE will fill in the rest. I'll know right away that I can set the `expectedResult` with `setExpectedResult`, or `setUriParts`, or `setVerifyYaml`. With an array I'd definitely have to look at it again. There's a good chance I'd have to look at it again between tests, too, if I didn't override something in one test but needed to do it in another.

rkaiser-dev (approved):

AricStewart:
Ok this is not actually ready for commit but an example.  If you like this builder pattern I will do some clean up and squashing to make it a clean patchset without the inside step

rkaiser-dev (approved):
I think this is definitely a step in a better direction with the addition of the builder pattern variant. Ideally we could get this to a place where `CiabTestCase` can keep `$unpriv_tokens` and `runRequest` as protected rather than `public` functions, but I think that's something that can evolve into the better picture over time.

rkaiser-dev (approved):

PR #673, opened 2024-01-14
Add PUT endpoint for departments

rkaiser-dev:
This PR adds the PUT endpoint for departments, including service and repository implementations, tests, and openapi documentation updates.

Also fixes an issue with PostDepartment openapi documentation schema.


api/src/App/Routes.php
```diff
@@ -35,6 +35,7 @@ function () use ($app, $authMiddleware) {
         function () use ($app, $authMiddleware) {
             $app->get('[/]', 'App\Controller\Department\ListDepartments');
             $app->post('[/]', 'App\Controller\Department\PostDepartment');
+            $app->put('/{id}', 'App\Controller\Department\PutDepartment');
```
AricStewart, line 38: I personally like the use if `{id}` but in quite a few other places we are using `{name}`.  I think the APIs will also accept a name or ID,  do we want to try to be consistant here or is that overkill?

api/src/Controller/Department/PutDepartment.php
```diff
@@ -0,0 +1,120 @@
...
+        $this->departmentService = $container->get("DepartmentService");
+        $this->includes = null;
+
+    }
+
+
+    public function buildResource(Request $request, Response $response, $params): array
+    {
+        $departmentId = $params["id"];
+        $body = $this->checkRequiredBody($request, []);
```
AricStewart, line 89: I do not this this is required with an empty set.

api/src/Controller/Department/PutDepartment.php
```diff
@@ -0,0 +1,120 @@
...
+            throw new InvalidParameterException("Name must be a non-empty string");
+        }
+
+        if (array_key_exists("ParentID", $body) && $body["ParentID"] <= 0) {
+            throw new InvalidParameterException("ParentID must be greater than 0");
+        }
+
+        if (array_key_exists("FallbackID", $body) && $body["FallbackID"] <= 0) {
+            throw new InvalidParameterException("FallbackID must be greater than 0");
+        }
```
AricStewart, line 105: What if they do a Put with no body. Then does just nothing change and we succeed?
Do we want to catch that case for simplicity?

api/src/App/Routes.php
rkaiser-dev: We opt for `{name}` here if we're looking for consistency. I'll check to see if anything actually passes in `{name}` to the other endpoints for department and make the switch if needed. Ideally we do ID based lookup only because name could be ambiguous.

api/src/Controller/Department/PutDepartment.php
rkaiser-dev: OK, will remove this.

api/src/Controller/Department/PutDepartment.php
rkaiser-dev: Yeah, that's a good question. I think we can handle that case here for sure.

api/src/App/Routes.php
rkaiser-dev: I took a look through the other endpoints that are using `{name}` syntax for Announcements and Deadlines. I poked around in the UI a bit and I confirmed through debugging that they are using the Department ID value. I know all of the ConCom List stuff should be using the Department ID value as well.

Is it worth considering the move to prefer ID for Department here and we can eventually update the other endpoints to ID numbers only?

rkaiser-dev:
This should be ready for another look. Left a comment asking about whether we move to ID for all Department endpoints and start with this one.

Also added a few more test cases to catch some things that the OpenAPI schema wouldn't normally allow, but that the server should still guard against and reject, assuming an all ID approach.

api/src/App/Routes.php
AricStewart: Lets make a low priority task to try to standardize our language and make {id} for all places we are using IDs.

api/src/Controller/Department/PutDepartment.php
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+
+    private function validateRequestParams(Request $request): array
+    {
+        $body = $request->getParsedBody();
+        if ($body == null || count($body) == 0) {
+            throw new InvalidParameterException("Must include at least one of 'Name', 'ParentID', or 'FallbackID'");
+        }
+
```
AricStewart, line 113: This does not quite do that.  I could include dummy things in the body `{'A': 'goo', 'B': 'gaa'}`  that would fool this check and not help.

AricStewart (approved):

PR #674, opened 2024-01-19
Aric/test revisions

AricStewart:


api/src/Tests/Cases/CycleTest.php
```diff
@@ -15,23 +16,19 @@ protected function setUp(): void
...
-            '/cycle',
-            null,
-            ['date_from' => $when, 'date_to' => $to],
-            201,
-            null,
-            '/cycle'
-        );
+        $this->cycle = testRun::testRun($this, 'POST', '/cycle')
+            ->setBody(['date_from' => $when, 'date_to' => $to])
+            ->run();
```
rkaiser-dev, line 21: Not an urgent thing, and this comment would apply in a similar way to `tearDown()` as well.

Should we interact directly with the database rather than via API call here for test setup/teardown? Although I imagine it'd be rare, this reliance on the API to create and delete objects for us to test against could result in test failures if the create or delete fails, I think.

api/src/Tests/Cases/CycleTest.php
```diff
@@ -40,25 +37,21 @@ protected function tearDown(): void
...
-            '/cycle',
-            null,
-            ['date_from' => $when, 'date_to' => $to],
-            201,
-            null,
-            '/cycle'
-        );
-
-        $data2 = $this->runSuccessJsonRequest('GET', '/cycle/'.$data->id, null, null, 200, null, '/cycle/{id}');
+        $data = testRun::testRun($this, 'POST', '/cycle')
```
rkaiser-dev, line 48: Is the plan to follow up with this and break this out into nice segmented tests as well?

rkaiser-dev (approved):

api/src/Tests/Cases/CycleTest.php
AricStewart: I debated that back and forth also. Doing actauly tests in setup/teardown seems odd but it also is  a nice clean path to do the work.  I decided on the side of having it a test just because but I am not sold either way. We can bring it up as we go forward. I know there are other tests where all these setup and such is done directly tot the database.

api/src/Tests/Cases/CycleTest.php
AricStewart: We can continue to break things up and refine them yes.  Always in the mode to keep expanding and improving tests.

PR #675, opened 2024-01-24
Add endpoint to delete departments

rkaiser-dev:
This PR adds an endpoint to Delete Departments/Divisions.

Currently there is no permission attached to this, but the original also lacked any sort of permission check on the back-end. I'm not super fond of this and would like to figure out a strategy for this going forward.

It also looks like maybe I need to include some deletes on `EMails` and `EmailListAccess` tables for this call as well, which is not currently implemented.

### DIFF
--- a/api/src/Repository/DepartmentRepository.php
+++ b/api/src/Repository/DepartmentRepository.php
@@ -2,7 +2,6 @@
 
 namespace App\Repository;
 
-use Exception;
 use Atlas\Query\Select;
 use Atlas\Query\Insert;
 use Atlas\Query\Update;
@@ -116,7 +115,24 @@ class DepartmentRepository implements RepositoryInterface
 
     public function deleteById(/*.mixed.*/$id): void
     {
-        throw new Exception(__CLASS__.": Method '__FUNCTION__' not implemented");
+        $placeholderDeptId = Select::new($this->db)
+            ->columns("DepartmentID")
+            ->from("Departments")
+            ->where("Name = 'Historical Placeholder'")
+            ->fetchOne();
+        
+        $update = Update::new($this->db);
+        $update->table("Departments")
+            ->column("FallbackID", null)
+            ->column("ParentDepartmentID", $placeholderDeptId["DepartmentID"])
+            ->whereEquals(["DepartmentID" => $id])
+            ->perform();
+
+        $update = Update::new($this->db);
+        $update->table("Departments")
+            ->column("FallbackID", null)
+            ->whereEquals(["FallbackID" => $id])
+            ->perform();
 
     }
 
--- a/api/src/App/Dependencies.php
+++ b/api/src/App/Dependencies.php
@@ -7,9 +7,11 @@ use Psr\Container\ContainerInterface;
 use App\Handler\ApiError;
 use App\Handler\RBAC;
 use App\Repository\DepartmentRepository;
+use App\Repository\EmailRepository;
 use App\Repository\EventRepository;
 use App\Repository\MemberRepository;
 use App\Service\DepartmentService;
+use App\Service\EmailService;
 use App\Service\EventService;
 use App\Service\MemberService;
 use Slim\App;
@@ -39,6 +41,12 @@ function setupAPIDependencies(App $app, array $settings)
 
     $container['RBAC'] = function (): RBAC {
         return new RBAC;
+    };
+
+    $container['EmailService'] = function ($c): EmailService {
+        return new EmailService(
+            new EmailRepository($c['db'])
+        );
     };
 
     $container['DepartmentService'] = function ($c): DepartmentService {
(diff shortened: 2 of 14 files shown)


api/src/Repository/DepartmentRepository.php
```diff
@@ -116,7 +117,23 @@ public function update(/*.string.*/$id, /*.mixed.*/$data): void
 
     public function deleteById(/*.mixed.*/$id): void
     {
-        throw new Exception(__CLASS__.": Method '__FUNCTION__' not implemented");
+        $placeholderDeptId = Select::new($this->db)
```
rkaiser-dev, line 120: Seems like I may need to add a few things here:

- Delete from EMails on DepartmentID
- Delete from EmailListAccess on DepartmentID
- Update other department FallbackIDs to null on DepartmentID

May be able to clean up the Placeholder Dept ID lookup

api/src/Repository/DepartmentRepository.php
```diff
@@ -116,7 +117,23 @@ public function update(/*.string.*/$id, /*.mixed.*/$data): void
...
+        $update->table("Departments")
+            ->column("FallbackID", null)
+            ->column("ParentDepartmentID", $placeholderDeptId["DepartmentID"])
+            ->whereEquals(["DepartmentID" => $id])
+            ->perform();
+
+        $delete = Delete::new($this->db);
+        $delete->from("Departments")
+            ->whereEquals(["DepartmentID" => $id])
+            ->perform();
```
AricStewart, line 137: So deleting departments has been a bit tricky.  This is because it breaks historical concom records and nobody is happy about that. 

if you look at `modules/concom/functions/concom.inc` and `delete_position` around line 336 you will see that we do not delete a position (department)  instead we set the Parent to be the magical 'Historical Placeholder' department.  That marks it in the system as a department that is no long active.  But it keeps all the dependancies in the concom records so that we do not loose any of the years served sort of logic.

api/src/Repository/DepartmentRepository.php
AricStewart: yeah, we will need to handle all those.  Just removing the emails and such should be fine. One thing the current code does not do cleanly is handle staff assigned to a department being deleted. Presently it is just operation procedure that there needs to be zero staff assigned to the department before it is deleted.  We could enforce this in code which may be good.

api/src/Repository/DepartmentRepository.php
rkaiser-dev: Sure, I can make that adjustment instead and opt to simply update it to use the historical as the parent.

It sounds like years served on ConCom is the only major requirement. If we had another approach for that would we be able to delete the departments? Or is the preference to keep with this forever?

api/src/Repository/DepartmentRepository.php
rkaiser-dev: I like the idea of enforcing that in code. I will make sure to update this to handle the other cases for now. I'll check on whether we can handle staff on a deleted department as well. 

Do the staff that are deleted from one department get added to other departments typically? Would we want to be able to keep track of years served?

AricStewart: Generally yes, they are moved to other departments.  But again it is a soft policy thing. I am sure there would be a case at some point where a staff member will leave or such.

And yes tracking years served is important.

api/src/Repository/DepartmentRepository.php
AricStewart: When I initially implemented this I asked if we wanted to track just years served or years served with what department and I never was given a clear answer.  Especially since they have a tendency to rename departments a lot, which already breaks that sort of history in some ways. But I just errored with the way that saved the most data possible for now.  It does mean the convention has many many historical departments, which is not a lot of data but still gets a bit messy feeling.

rkaiser-dev:
I've added Email and EmailListAccess Service/Repository implementations since they are also needed here to fulfill the logic that was in use before.

I've updated DepartmentService to use these and DepartmentRepository to not actually delete the department.

This should be ready for another look.

api/src/Repository/DepartmentRepository.php
```diff
@@ -6,6 +6,7 @@
 use Atlas\Query\Select;
 use Atlas\Query\Insert;
 use Atlas\Query\Update;
+use Atlas\Query\Delete;
```
AricStewart, line 9: This is not required anymore.

AricStewart (changes requested):
One tiny cleanup then all good.

api/src/Repository/DepartmentRepository.php
rkaiser-dev: All set. Removed `use Exception` as well since it was not used either.

AricStewart (approved):

PR #681, opened 2024-02-17
Update sidebar with ability to add departments

rkaiser-dev:
This PR adds the ability to create new divisions and departments on the ConCom Structure Page.

We also gain the ability to switch out the sidebar view, starting with emails, without losing the department data we have filled out, and without using multiple rendered sidebars.

It demonstrates how composables allow us to keep our components clean. We will be able to more easily test our components because we can now mock the composable functions.

### DIFF
--- a/modules/concom/vue/views/StaffStructurePage.js
+++ b/modules/concom/vue/views/StaffStructurePage.js
@@ -1,6 +1,8 @@
 /* globals Vue, showSpinner, hideSpinner */
 import { useCurrentUser } from '../composables/currentUser.js';
 import { useDivisionHierarchy } from '../composables/divisionHierarchy.js';
+import { useSidebar } from '../composables/sidebar.js';
+import { useSidebarActions } from '../composables/sidebarActions.js';
 
 const TEMPLATE = `
   <div class="UI-container">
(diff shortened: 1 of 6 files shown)


modules/concom/vue/views/StaffStructurePage.js
```diff
@@ -1,13 +1,15 @@
...
+import { useSidebar } from '../composables/sidebar.js';
+import { useSidebarActions } from '../composables/sidebarActions.js';
 
 const TEMPLATE = `
   <div class="UI-container">
     <div :class="sidebarDisplayClass">
       <div class="CONCOM-admin-content">
         <div class="CONCOM-admin-section">
-          <span>ConCom Structure</span>
+          <span>ConCom Structure (NEW)</span>
```
rkaiser-dev, line 12: I will fix this in a future PR. I am currently using this while testing and committed it by mistake.

modules/concom/vue/views/StaffStructurePage.js
```diff
@@ -61,12 +63,17 @@ function setup() {
...
+    closeSidebar,
+    saveSidebarData,
     divisions,
     canEditRbac,
     showSidebar,
     sidebarName,
     sidebarData,
-    sidebarDisplayClass
+    sidebarDisplayClass,
+    sidebarTempData: []
```
rkaiser-dev, line 76: I will remove `sidebarTempData` in a future commit, I originally had a plan for this before moving logic to composables.

PR #704, opened 2024-05-01
Fix concom modification by non-admin staff

AricStewart:


api/src/Controller/BaseController.php
```diff
@@ -138,6 +138,11 @@ abstract class BaseController
     */
     protected $currentEvent = -1;
 
+    /**
+     * @var bool
+    */
+    public $internal = false;
```
rkaiser-dev, line 144: What are some other things that `$internal` would apply to? With members I can see it was in the FindMembers controller, are there any other immediate examples?

AricStewart: Not that I know about here.  But moving it into here makes it easier to set in `IncludeResource` and check in `GetMember` Otherwise we have to see if the property exists before setting it or checking it and it just felt cleaner to move it into the Base.

rkaiser-dev (approved):

PR #708, opened 2024-07-09
Fix and simplify the volunteer hour entry field

AricStewart:


modules/volunteers/admin/vue/hourEntryField.js
```diff
@@ -38,28 +28,22 @@ export default {
...
-        if (this.enterHours != null) {
-          const v = this.enterHours;
-          this.enterHours = null;
-          return v;
-        }
         return Math.floor(this.modelValue);
       },
       set(value) {
-        this.enterHours = Math.floor(value);
+        var enterHours = Math.floor(value);
```
rkaiser-dev, line 34:
```suggestion
        const enterHours = Math.floor(value);
```

modules/volunteers/admin/vue/hourEntryField.js
```diff
@@ -38,28 +28,22 @@ export default {
...
-        if (this.enterMin != null) {
-          const v = this.enterMin;
-          this.enterMin = null;
-          return v;
-        }
         return Math.floor((this.modelValue - Math.floor(this.modelValue)) * 60);
       },
       set(value) {
-        this.enterMin = Math.floor(value);
+        var enterMin = Math.floor(value);
```
rkaiser-dev, line 44:
```suggestion
        const enterMin = Math.floor(value);
```

modules/volunteers/admin/vue/hourEntryField.js
```diff
@@ -38,28 +28,22 @@ export default {
...
-          this.enterHours = null;
-          return v;
-        }
         return Math.floor(this.modelValue);
       },
       set(value) {
-        this.enterHours = Math.floor(value);
+        var enterHours = Math.floor(value);
+        value = enterHours + (this.modelValue % 1)
+        this.$emit('update:modelValue', value);
```
rkaiser-dev, line 36: It looks like changing `value` should automatically result in an emit call, right? Or is it not triggering `computed.value.set()`?

AricStewart: Maybe... It has been forever since I authored this and I do not recall seeing it happen but I may have missed it.

rkaiser-dev (approved):

